Issue #27, opened 2026-08-05
T25 — Platform-conditional helper: FDA probe, restic discovery in Core


bherila:
# T25 — Platform-conditional helper: FDA probe, restic discovery in Core

**Size:** M · **Model:** Sonnet · **Depends on:** T21 · **Milestone:** M5

## Repository context (read first)
Two macOS-specific behaviours currently live in the wrong place for a Linux build.

**1. Full Disk Access.** `Helper/Sources/Commands/FdaCheck.swift` probes `~/Library/Safari` (falling back to `~/Library/Mail`) and writes `state/fda-check.json`, recording the `context` the probe ran in (`launchd` by default). This exists because macOS silently denies access to Mail/Safari/Messages data without FDA — no prompt, just empty backups. See `docs/keychain-and-fda.md` §2. On Linux the whole concept is meaningless: there is no TCC, and ordinary file permissions govern access.

**2. restic discovery.** `App/Sources/Support/ResticDiscovery.swift` finds a usable restic binary by probing well-known locations (`/opt/homebrew/bin/restic`, …) and then `PATH`, validating each candidate by actually running `restic version --json` and comparing the version — a candidate counts as found only if it *ran*, not merely if it exists and is `+x`. It has a per-candidate timeout and a cap on candidates examined so a pathological `PATH` cannot stall onboarding, and the search paths are injectable. It is well-built, but it lives in `App/`, which is macOS-only and never compiled on Linux. On Linux, `HelperContext.make()` currently exits 1 with "restic not configured — open Restic Station" — advice that is impossible to follow on a headless host.

## Goal
The helper behaves sensibly on Linux: it finds restic on its own, and it does not pretend a macOS privacy mechanism exists.

### 1. Move `ResticDiscovery` to Core
- Relocate to `Core/Sources/ResticStationCore/Restic/ResticDiscovery.swift`, public, with its injectable search paths and its existing safety properties preserved exactly — the probe-by-running rule, the per-candidate timeout, and the candidate cap are all load-bearing; do not "simplify" them.
- Make the well-known-location list platform-conditional: macOS keeps `/opt/homebrew/bin/restic` and the current list; Linux uses `/usr/bin/restic`, `/usr/local/bin/restic`, `/opt/restic/bin/restic`. `PATH` scanning stays on both.
- Update `App/Sources/Support/` to import it from Core rather than define it. The app's onboarding behaviour must not change; this is a move, not a redesign.
[...]

PR #34, opened 2026-08-05
T22 — Portable AppPaths (XDG on Linux)

### LINKED ISSUE #24: T22 — Portable AppPaths (XDG on Linux)
# T22 — Portable AppPaths (XDG on Linux)

**Size:** S · **Model:** Sonnet · **Depends on:** T21 · **Milestone:** M5

`AppPaths` (`Core/Sources/ResticStationCore/Config/AppPaths.swift`) is the single source of truth for every runtime path Restic Station reads or writes — `config.json`, `runs/`, `state/`, `locks/`, `mounts/`, and restic's cache dir. Nothing else in the codebase may hard-code a path. `docs/architecture.md` §AppPaths carries the normative table this file mirrors.

Today two things are macOS-specific:
- `AppPaths.default()` (line ~24) builds `~/Library/Application Support/ResticStation` from `FileManager.default.homeDirectoryForCurrentUser`.
- `resticCacheDir` (line ~120) is fixed at `~/Library/Caches/net.herila.ResticStation/restic`, deliberately independent of `root` because it is a regenerable cache rather than app state.

There is already an escape hatch: `RESTIC_STATION_DATA_DIR`, honoured by `default()` and used by the test suite and `scripts/integration-test.sh`.

## Goal
`AppPaths.default()` returns correct, conventional locations on Linux without changing anything on macOS, so a headless helper on Linux stores its state where a Linux admin expects to find it.

## Do
1. **Branch `default()` on `#if os(Linux)`** with this precedence, checked in order:
   1. `RESTIC_STATION_DATA_DIR` if set and non-empty (unchanged, both platforms).
   2. Linux: `$XDG_STATE_HOME/restic-station`, falling back to `~/.local/state/restic-station` when `XDG_STATE_HOME` is unset or empty. State — not config — is the right XDG base dir here: `config.json` is the smallest part of `root`, which is dominated by `runs/`, `state/`, and `locks/`.
   3. macOS: `~/Library/Application Support/ResticStation` (unchanged).
   Per the XDG spec, ignore an `XDG_*` value that is not an absolute path and fall back as if it were unset.
2. **Branch `resticCacheDir`** the same way: Linux uses `$XDG_CACHE_HOME/restic-station/restic` → `~/.cache/restic-station/restic`; macOS keeps `~/Library/Caches/net.herila.ResticStation/restic`.
3. **Keep every other member platform-agnostic.** `configFile`, `runsDir`, `stateDir`, `locksDir`, `mountsDir`, and the per-id helpers are all `root`-relative already — they must stay untouched so the on-disk layout below `root` is byte-identical across platforms. This matters: T27's `config export`/`import` and any user rsync-ing a directory between hosts depend on it.
4. **Update `docs/architecture.md` §AppPaths** with a two-column macOS/Linux table, and fix the doc comments at the top of `AppPaths.swift` which currently state the macOS path as if unconditional.
5. Note in the doc that `mountsDir` is macOS-only in practice (`restic mount` needs macFUSE there, FUSE on Linux) but the path itself stays defined on both.

## Tests / verification
`Core/Tests/ResticStationCoreTests/Config/AppPathsTests.swift` currently asserts the macOS layout; T21 gated it for Linux. Ungate it and restructure:
- Shared assertions for the `root`-relative layout, run on both platforms against an explicit `AppPaths(root:)` — these are the ones that guarantee cross-platform layout identity.
- `RESTIC_STATION_DATA_DIR` override honoured on both platforms; a set-but-empty value falls through to the platform default.
- Platform-gated assertions for what `default()` resolves to, including: `XDG_STATE_HOME` set → used; unset → `~/.local/state/restic-station`; set to a relative path → ignored, falls back.
- Add a test asserting that two `AppPaths` with the same `root` produce identical relative sub-paths regardless of platform (guards requirement 3 against future drift).

Environment-dependent tests must set and restore env vars around themselves rather than relying on ambient state.

## Acceptance criteria
- [ ] `swift test` green on macOS and on Linux (`swift:6.0`), with the T21 gate on `AppPathsTests` removed.
- [ ] macOS resolution is byte-identical to before this change — verified by a test, not by inspection.
- [ ] Everything below `root` is identical across platforms.
- [ ] `docs/architecture.md` §AppPaths documents both platforms; `AppPaths.swift` doc comments no longer claim an unconditional `~/Library` path.

---

**Blocked by:** #23

bherila:
## What changed

`AppPaths` is now portable. Only two members branch on the OS; everything else is untouched.

**`Core/Sources/ResticStationCore/Config/AppPaths.swift`**

- `default()` resolution order, checked in this order on both platforms:
  1. `RESTIC_STATION_DATA_DIR` if set and non-empty (unchanged).
  2. Linux: `$XDG_STATE_HOME/restic-station`, falling back to `~/.local/state/restic-station`.
  3. macOS: `~/Library/Application Support/ResticStation` (unchanged).
- `resticCacheDir`: Linux `$XDG_CACHE_HOME/restic-station/restic` → `~/.cache/restic-station/restic`; macOS keeps `~/Library/Caches/net.herila.ResticStation/restic`. Still deliberately independent of `root`.
- New internal `xdgBaseDirectory(variable:fallbackHomeRelativeComponents:)` (Linux-only), which honours an `XDG_*` value only when it is set, non-empty **and** absolute — a relative value is ignored and the fallback applies, per the XDG Base Directory Specification.
- `configFile`, `runsDir`, `stateDir`, `locksDir`, `mountsDir`, and every per-id helper are unchanged and remain purely `root`-relative.
- Doc comments no longer claim an unconditional `~/Library` path; `mountsDir` gained a note that it is macOS-only in practice (macFUSE / FUSE) though the path stays defined on both.

**`Core/Tests/ResticStationCoreTests/Config/AppPathsTests.swift`** — restructured:

- A `withEnvironment(_:_:_:)` helper sets and restores (via `defer`, so also on throw) any env var it touches, so nothing leaks between tests. All env-touching tests live in the `.serialized` `AppPathsEnvTests` suite — `resticCacheDirIsIndependentOfRoot` moved there because on Linux it now reads `XDG_CACHE_HOME`.
- Cross-platform: `RESTIC_STATION_DATA_DIR` honoured; set-but-empty falls through to the platform default; the override beats `XDG_STATE_HOME`.
- Cross-platform layout: `computedPathsMatchArchitectureTable` now also covers `runsDir`/`runDir`/`stateDir`/`locksDir`. New `subPathsAreRelativeToRootAndPlatformIndependent` builds `AppPaths` from two different roots, strips each root prefix, asserts the two relative lists are equal **and** equal to a literal expected list — so a platform branch sneaking in below `root` fails the test on both OSes.
- macOS-gated (`#if os(macOS)`): `macOSDefaultRootIsUnchanged` and `macOSResticCacheDirIsUnchanged` pin the exact pre-change absolute strings, and `macOSIgnoresXDGVariables` asserts `XDG_STATE_HOME`/`XDG_CACHE_HOME` have no effect on macOS.
- Linux-gated (`#if os(Linux)`): `XDG_STATE_HOME` absolute → used; unset → `~/.local/state/restic-station`; empty → fallback; relative → ignored, fallback. Same four-way coverage for `XDG_CACHE_HOME` (absolute / unset / relative).

**`docs/architecture.md` §AppPaths** — replaced the single-platform sentence with a two-column macOS/Linux table for `root` and the restic cache, the rationale for choosing the *state* XDG base dir over *config*, the absolute-path rule for `XDG_*`, an explicit statement that everything below `root` is byte-identical across platforms (with a test guaranteeing it), and a `mounts/<destId>/` row plus the macFUSE/FUSE caveat.

## How verified

Locally on macOS (Apple Swift 6.3.3):

- `swift build` — clean
- `swift test` (root package) — pass
- `swift test --package-path Core` — 313 tests in 52 suites pass
- `./scripts/bootstrap.sh && xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO -quiet` — exit 0

Linux verification is CI-only (`ubuntu-24.04-arm`, container `swift:6.1`) — no local Linux toolchain available in this environment.

## Acceptance criteria

- [x] `swift test` green on macOS (verified locally) and on Linux (verified in CI — see the `linux` job). Note: the issue's premise that T21 gated `AppPathsTests` for Linux was wrong; T21 added no `#if os(...)` gates, so there was no gate to remove. The old assertions passed on Linux only because the pre-change implementation was unconditionally `~/Library/...` on every platform.
- [x] macOS resolution byte-identical to before — verified by test, not inspection (`macOSDefaultRootIsUnchanged`, `macOSResticCacheDirIsUnchanged`, `macOSIgnoresXDGVariables` pin the literal strings).
- [x] Everything below `root` identical across platforms — verified by `subPathsAreRelativeToRootAndPlatformIndependent`, which runs on both platforms and asserts against literal relative paths.
- [x] `docs/architecture.md` §AppPaths documents both platforms; `AppPaths.swift` doc comments no longer claim an unconditional `~/Library` path.

## Deferred / out of scope

- `docs/restic-cli.md` still states `RESTIC_CACHE_DIR=~/Library/Caches/net.herila.ResticStation/restic` and the `mount` mountpoint as unconditional macOS paths. Left alone deliberately: the issue names only `docs/architecture.md`, and that file is being edited concurrently by the restic-discovery task (#27). Worth a one-line follow-up once that lands.
- The XDG runtime dir is not used for `locks/`. `locks/` stays under `root` because the lock files are part of the byte-identical-below-`root` contract; revisit only if lock semantics change.

Closes #24

### DIFF
--- a/Core/Sources/ResticStationCore/Config/AppPaths.swift
+++ b/Core/Sources/ResticStationCore/Config/AppPaths.swift
@@ -3,9 +3,18 @@ import Foundation
 /// Single source of truth for every runtime path Restic Station reads or
 /// writes. Never hard-code a path elsewhere — go through `AppPaths`.
 ///
-/// `root` defaults to `~/Library/Application Support/ResticStation` and is
-/// overridable via `init(root:)` and via the `RESTIC_STATION_DATA_DIR`
-/// environment variable (used by tests and `scripts/integration-test.sh`).
+/// `root` is platform-dependent:
+/// - macOS: `~/Library/Application Support/ResticStation`
+/// - Linux: `$XDG_STATE_HOME/restic-station`, falling back to
+///   `~/.local/state/restic-station`
+///
+/// It is overridable via `init(root:)` and via the `RESTIC_STATION_DATA_DIR`
+/// environment variable (used by tests and `scripts/integration-test.sh`),
+/// which takes precedence on both platforms.
+///
+/// **Everything below `root` is byte-identical across platforms** — only
+/// `root` and `resticCacheDir` branch on the OS. `config export`/`import`
+/// and rsync-ing a data directory between hosts depend on this.
 ///
 /// See `docs/architecture.md` §AppPaths for the full table this mirrors.
 public struct AppPaths: Equatable, Sendable {
@@ -16,17 +25,45 @@ public struct AppPaths: Equatable, Sendable {
     }
 
     /// Resolves the root from `RESTIC_STATION_DATA_DIR` if set (non-empty),
-    /// else `~/Library/Application Support/ResticStation`.
+    /// else the platform default: `~/Library/Application Support/ResticStation`
+    /// on macOS, `$XDG_STATE_HOME/restic-station` (or
+    /// `~/.local/state/restic-station`) on Linux.
     public static func `default`() -> AppPaths {
         if let override = ProcessInfo.processInfo.environment["RESTIC_STATION_DATA_DIR"], !override.isEmpty {
             return AppPaths(root: URL(fileURLWithPath: override, isDirectory: true))
         }
+        #if os(Linux)
+        return AppPaths(root: xdgBaseDirectory(
+            variable: "XDG_STATE_HOME",
+            fallbackHomeRelativeComponents: [".local", "state"]
+        ).appendingPathComponent("restic-station", isDirectory: true))
+        #else
         let appSupport = FileManager.default.homeDirectoryForCurrentUser
             .appendingPathComponent("Library", isDirectory: true)
             .appendingPathComponent("Application Support", isDirectory: true)
             .appendingPathComponent("ResticStation", isDirectory: true)
         return AppPaths(root: appSupport)
+        #endif
     }
+
+    #if os(Linux)
+    /// Resolves an XDG base directory per the XDG Base Directory Specification:
+    /// the environment variable is honoured only when set, non-empty, **and**
+    /// an absolute path; anything else falls back to the home-relative default
+    /// as if the variable were unset.
+    static func xdgBaseDirectory(variable: String, fallbackHomeRelativeComponents: [String]) -> URL {
+        if let value = ProcessInfo.processInfo.environment[variable],
+           !value.isEmpty,
+           value.hasPrefix("/") {
+            return URL(fileURLWithPath: value, isDirectory: true)
+        }
+        var url = FileManager.default.homeDirectoryForCurrentUser
+        for component in fallbackHomeRelativeComponents {
+            url = url.appendingPathComponent(component, isDirectory: true)
+        }
+        return url
+    }
+    #endif
 
     // MARK: - config.json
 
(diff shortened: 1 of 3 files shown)


bherila:
@codex review

Issue #26, opened 2026-08-05
T24 — Per-machine config scoping (schema v2) + migration


bherila:
# T24 — Per-machine config scoping (schema v2) + migration

**Size:** L · **Model:** Opus · **Depends on:** T22 · **Milestone:** M5
**Safety-critical — requires Fable review before merge.** A resolution bug here silently backs up the wrong directories, or silently backs up nothing.

## Repository context (read first)
`docs/data-model.md` is normative and must be updated as part of this task.

`config.json` is already fully portable in *format* — plain `Codable` JSON, atomically written by `ConfigStore` (`Core/Sources/ResticStationCore/Config/ConfigStore.swift`), no plists, no security-scoped bookmarks, no binary blobs. `ConfigStore.load()` already rejects a config whose `version` exceeds `AppConfig.currentVersion` with `ConfigError.newerVersion`, so an older build meeting a v2 file fails cleanly rather than misreading it.

Current schema (`Core/Sources/ResticStationCore/Config/Models.swift`):
```
AppConfig     version, resticPath?, showMenuBarIcon, onboardingCompleted?, sets[]
BackupSet     id, name, sources[], excludes[], schedule, retention?, checkPolicy?,
              stalenessWarningDays, destinations[]
Destination   id, label, repoURL, isPrimary, nonSecretEnv
```

The problem: a config authored in the macOS GUI holds `/Users/bwh/Documents` source paths, an `/opt/homebrew/bin/restic` binary path, and possibly a `/Volumes/Big/repo` destination — none of which exist on a Linux host. The format ports; the *values* do not.

**Decided design (do not redesign this):** full parity — a Linux host may be either a backup source with its own directories, or a mirror/restore target that never reads local sources. One `config.json` is shared across every machine; each machine additionally has a small, never-shared `machine.json`.

Explicitly rejected: automatic path rewriting / prefix mapping (`/Users/bwh` → `/home/bwh`). It is implicit, silently wrong at the edges, and backups are the wrong place for magic. Per-machine overrides are explicit.

### 1. `machine.json` — host-local, never shared
New file at `<AppPaths.root>/machine.json`, loaded by a new `MachineStore` alongside `ConfigStore`:
```json
{ "version": 1, "machineId": "studio-mac", "resticPath": "/usr/bin/restic" }
```
[...]

Issue #25, opened 2026-08-05
T23 — SecretStore abstraction + Linux file backend + secret CLI


bherila:
# T23 — SecretStore abstraction + Linux file backend + `secret` CLI

**Size:** L · **Model:** Opus · **Depends on:** T21 · **Milestone:** M5
**Safety-critical — requires Fable review before merge.** This task handles repository passwords. A bug here either loses access to backups or leaks credentials.

## Repository context (read first)
`docs/keychain-and-fda.md` is normative for the macOS half and must be read in full before starting.

`KeychainClient` (`Core/Sources/ResticStationCore/Secrets/KeychainClient.swift`) is a `struct` holding a `ProcessRunning`, with this surface:

```
setPassword(_:destId:) / password(destId:) / deletePassword(destId:)
setSecretEnv(_:destId:) / secretEnv(destId:) / deleteSecretEnv(destId:)
static passwordCommand(destId:) -> String
```

Every operation shells out to `/usr/bin/security` — never `SecItemAdd` — because items created with `-T /usr/bin/security` are the only ones a headless launchd context can read without a GUI consent prompt. Storage keys are the destination UUID lowercased, and `"<uuid>-env"` for the secret-env JSON blob. `/usr/bin/security` does not exist on Linux; this is the single largest blocker in M5.

Consumers today: `ResticRunner` (`Core/.../Restic/ResticRunner.swift`) holds a concrete `KeychainClient` (line ~85), does a keychain pre-flight read so a locked keychain becomes a clean retryable `ResticRunnerError.keychainUnavailable` rather than a confusing restic failure (~line 153), and sets `RESTIC_PASSWORD_COMMAND` / `RESTIC_FROM_PASSWORD_COMMAND` from `KeychainClient.passwordCommand(destId:)` (~lines 177–179). `HelperContext` constructs it (`Helper/Sources/HelperContext.swift:77`). The app constructs it via `App/Sources/Support/`.

## Goal
Abstract secret storage behind a protocol; keep the macOS keychain backend behaviourally identical; add a Linux file backend that works with no desktop session, no keyring daemon, and no GPG agent; and add CLI subcommands to manage secrets, because the Linux host has no GUI to enter a password into.

### 1. `SecretStore` protocol (`Core/Sources/ResticStationCore/Secrets/`)
Extract the six instance methods above into a `public protocol SecretStore: Sendable`, `async throws`, keyed by `destId: UUID`. Preserve the existing semantics exactly, they are load-bearing:
[...]

PR #35, opened 2026-08-05
T25 — Platform-conditional helper: FDA probe, restic discovery in Core

### LINKED ISSUE #27: T25 — Platform-conditional helper: FDA probe, restic discovery in Core

bherila:
### 1. `ResticDiscovery` moved to Core
- `App/Sources/Support/ResticDiscovery.swift` → `Core/Sources/ResticStationCore/Restic/ResticDiscovery.swift`, made `public`. Git records it as a rename (70% similarity).
- **All three safety properties survive unchanged**, and each now has an explicit test:
  - *probe-by-running*: a candidate counts as found only if `restic version --json` actually ran and parsed. Existence + `+x` is a filter, never the answer.
  - *per-candidate timeout* (5 s), asserted on every recorded call.
  - *candidate cap* (24), asserted both on the candidate list and on how many were actually executed.
- Well-known locations are platform-conditional and injectable:

  | Platform | List |
  | --- | --- |
  | macOS | `/opt/homebrew/bin/restic`, `/usr/local/bin/restic`, `/opt/local/bin/restic` |
  | Linux | `/usr/bin/restic`, `/usr/local/bin/restic`, `/opt/restic/bin/restic` |

  `PATH` scanning is unchanged on both. `ResticDiscovery.wellKnownPaths` (which `ResticSettings` uses to seed the open panel) still resolves to the current platform's list, so the app is unaffected.
- `App/` keeps only the *UI vocabulary*: `App/Sources/Support/ResticStatusMapping.swift` holds the `ResticProbe.status` / `ResticDiscoveryResult.status` → `ResticStatus` mapping as extensions on the Core types. `ResticStatus` stays in `AppModel.swift` — Core has no business knowing about it. This is a move, not a redesign; onboarding behaviour is unchanged.
- `ResticDiscoveryResult.status` used an inline `first { if case .tooOld … }`; that predicate is now `firstTooOld` on the Core type so the App extension stays a pure mapping. Same result.

### 2. Helper resolves restic itself
`Helper/Sources/ResticPathResolution.swift` (new; kept out of `HelperContext.swift` so that file's diff stays a two-line change — see the seam note below).

Resolution order today: **`AppConfig.resticPath` (deprecated) → discovery.** A discovered path is logged once at info level to stderr and deliberately **not** written back into `config.json`, which is shared across machines.

`HelperContext.make()` / `makeTolerant(…)` became `async` (every call site was already in an `async` command body).

When nothing resolves:
- macOS: the T10 wording verbatim — `restic not configured — open Restic Station`.
- Linux: names what was searched and how to fix it, e.g.
  ```
  restic not found. Searched /usr/bin/restic, /usr/local/bin/restic, /opt/restic/bin/restic, and every directory on PATH.
  Install restic (for example `apt install restic` or `dnf install restic`), or set "resticPath" in <root>/config.json.
  ```

### 3. `fda-check` is platform-aware
- macOS: byte-identical. Same probe (`~/Library/Safari` → `~/Library/Mail`), same `state/fda-check.json` schema, same `--context` flag, same always-exit-0 contract. Verified by running the built helper and comparing the written file against `docs/data-model.md`.
- Linux: still registered as a subcommand (scripts and docs stay uniform), reports "not applicable on this platform", exits 0, and **writes nothing**. `probeAndRecord` now returns `FdaCheckResult?` — `nil` off macOS.
- Absent-file semantics are now a Core rule rather than an App comment: `HealthDerivation.fullDiskAccessDenied(from:)` — `nil` → `false`; only an explicit `hasFullDiskAccess == false` is a denial. `AppModel` feeds `appHealth(…)` through it. Tested in Core, so the Linux suite holds the rule too.

### 4. Docs
- `docs/architecture.md`: new **§restic discovery** (the section the source already cited but which did not exist) with the per-platform table, the three rules, and the app-vs-helper resolution split; Core's component blurb now lists restic discovery.
- `docs/keychain-and-fda.md` §2: states FDA is macOS-only, documents `fda-check`'s Linux no-op and why a synthetic "granted" record was rejected, and adds a normative **Absent-file semantics** subsection.
- `docs/data-model.md` §state/fda-check.json: absence is meaningful; points at the normative rule.

## How verified

**Locally (macOS, Swift 6.3.3):**
- `swift build` — clean.
- `swift test` (root) — 8 tests, green.
- `swift test --package-path Core` — 328 tests in 53 suites, green.
- `./scripts/bootstrap.sh && xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO -quiet` — green (a type moved out of `App/`, so this was the load-bearing check).
- Ran the built helper: `fda-check --context smoke` writes the unchanged 4-key schema; `probe-repo` with `PATH=/nonexistent` and `resticPath: null` logs `info: restic not configured; discovered /opt/homebrew/bin/restic (version 0.18.1)` and proceeds past resolution.

**In CI (Linux, the only Linux available — no local toolchain, no container runtime):** three new steps in the `linux` job:
1. `fda-check` prints "not applicable on this platform" and `state/fda-check.json` does not exist afterwards.
2. With no restic anywhere, `probe-repo` fails with the actionable message naming `/usr/bin/restic` and the config path, and *not* "open Restic Station". With a `restic` on `PATH` that answers `version --json`, the same command logs `discovered <path>` and gets through to "no backup set with id …" — i.e. an unconfigured headless host resolves restic on its own.
3. `tick` exits 0.

The new `ResticDiscovery` tests inject both the well-known list and `PATH`, and use real `+x` files in a temp directory whose fate a path-keyed fake `ProcessRunning` decides — so **macOS CI genuinely exercises the Linux code path**, and no test depends on what is installed on the runner.

### The probe-by-running rule
This is the property the issue flags as most likely to be lost in a move, so it is asserted directly rather than implied: `executableThatCannotRunIsRejected` first asserts the filesystem answers *yes* to both cheap questions (`fileExists`, `isExecutableFile`), then scripts the runner to fail to launch, then asserts `chosen == nil`, the candidate is `.unusable`, **and** that it was rejected by having been run (`argv == [path, "version", "--json"]`). Sibling tests cover the other two ways a candidate can run and still not count: exits nonzero, and prints something that is not restic's version JSON.

## `TODO(#26)` seam

`machine.json` does not exist yet (#26 / T24 owns it), so the resolution order implemented is `AppConfig.resticPath → discovery`. The `machine.json` step is a one-line insertion at a marked point in `Helper/Sources/ResticPathResolution.swift`:

```swift
// TODO(#26): machine.json `resticPath` is checked here, ahead of the
// deprecated config.json fallback below. One line:
//     if let machinePath = machine.resticPath, !machinePath.isEmpty { return machinePath }
```

A second `TODO(#26)` in `resticNotFoundMessage(paths:discovery:)` marks where the Linux message should point at `machine.json` instead of `config.json` — a per-machine override is the right place for a per-machine binary path, but naming a file that does not exist yet would be misleading advice today.

## Darwin-assumption sweep (reported, not fixed)

Grepped `Core/Sources` and `Helper/Sources` for `os(macOS)`, `canImport(Darwin)`, AppKit/ServiceManagement/Security imports, `~/Library`, `/usr/bin/`, `launchctl`, `NSWorkspace`, `Bundle.main`, `homeDirectoryForCurrentUser`, `DistributedNotification`.

**Confirmed clean:** `Reachability` (Foundation only); `StateStore.swift` ~line 356 (`DistributedNotificationCenter` already `#if os(macOS)`); `ProcessRunning`, `FileLock`, `LogWriter`, `RunStore`, `ConfigStore` (all already `canImport(Darwin)`/`Glibc`-conditional).

**Still Darwin-assuming, owned by other in-flight work — not touched:**
1. `Core/Sources/ResticStationCore/Config/AppPaths.swift` — hardcodes `~/Library/Application Support/ResticStation` and `~/Library/Caches/net.herila.ResticStation/restic`. It compiles and runs on Linux (it is just a path), but the location is wrong there; XDG dirs are the Linux convention. **Owned by #24/T22 right now**, and the task brief explicitly barred editing it.
2. `Core/Sources/ResticStationCore/Secrets/KeychainClient.swift` — hardcodes `/usr/bin/security`, which does not exist on Linux. It compiles, but every secret read fails at runtime, which means no backup can run. This is the single biggest remaining blocker for a working Linux backup. **Owned by #25/T23 (SecretStore) right now**, and likewise barred.

**New finding, unowned:**
3. `Core/Sources/ResticStationCore/Support/HelperCommand.swift` — `LaunchctlCommand.executablePath = "/bin/launchctl"` and the `kickstart` argv builders are compiled into Core on every platform. Harmless today (pure argv construction; only the macOS app ever executes it, and nothing in `Helper/` calls it), but it is macOS-only vocabulary sitting in the portable layer. When a systemd equivalent is needed, this is where the platform split belongs. Flagged rather than fixed — not in this task's scope, and changing Core's public surface would collide with the two branches above.

## Acceptance criteria

- [x] **`swift test` green on macOS** — verified locally (root: 8 tests; Core: 328 tests) and in the `macos` CI job.
- [x] **`swift test` green on Linux** — verified in the `linux` CI job (root `swift build` + `swift test` + `swift test --package-path Core`). This is the only Linux verification available; there is no local Linux toolchain or container runtime.
- [x] **The macOS app builds and its onboarding restic-discovery behaviour is unchanged** — build verified locally and in CI. "Unchanged" is verified structurally (the discovery logic is the same code, the app's call sites are untouched apart from the import boundary, and `ResticDiscovery.wellKnownPaths` still yields the macOS list) plus the moved unit tests. The onboarding *sheet* itself is not exercised by an automated UI test in this repo — that remains manual-checklist territory, as before this PR.
- [x] **A Linux host with `restic` on `PATH` and no `resticPath` configured runs the helper successfully** — verified in CI via the discovery step: with only a `PATH` restic, the helper resolves it, logs it, and proceeds past resolution. Note the CI proof runs through `probe-repo` rather than `tick`, because a `tick` that reaches a *set* would need a valid set and would then hit the `/usr/bin/security` keychain blocker above (finding 2); `tick` shares the identical `resolveResticPath` code path via `makeTolerant`, and CI asserts `tick` exits 0. Full `tick`-with-a-set on Linux is **deferred to T23** for that reason.
- [x] **`docs/keychain-and-fda.md` states FDA is macOS-only and documents the absent-file semantics; `docs/architecture.md` reflects `ResticDiscovery` living in Core** — both updated, plus `docs/data-model.md`.

Closes #27

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,63 @@ jobs:
       - name: Run Core unit tests
         run: swift test --package-path Core
 
+      # T25 acceptance criteria that only a real Linux run can show.
+      - name: fda-check is a no-op on Linux
+        run: |
+          export RESTIC_STATION_DATA_DIR=$(mktemp -d)
+          OUT=$(.build/debug/restic-station-helper fda-check)
+          echo "$OUT"
+          echo "$OUT" | grep -q "not applicable on this platform"
+          # A fake "granted" record would make the file's meaning
+          # platform-dependent — nothing may be written at all.
+          test ! -e "$RESTIC_STATION_DATA_DIR/state/fda-check.json"
+
+      - name: restic discovery resolves the helper's binary on Linux
+        run: |
+          export RESTIC_STATION_DATA_DIR=$(mktemp -d)
+          mkdir -p "$RESTIC_STATION_DATA_DIR"
+          # No resticPath configured and no sets: enough to exercise
+          # HelperContext.make()'s resolution, which runs before anything else.
+          echo '{"version":1,"resticPath":null,"showMenuBarIcon":true,"sets":[]}' \
+            > "$RESTIC_STATION_DATA_DIR/config.json"
+          UUID=00000000-0000-0000-0000-000000000001
+
+          # 1. Nothing installed anywhere: an actionable message, not "open
+          #    Restic Station".
+          OUT=$(env PATH=/nonexistent .build/debug/restic-station-helper \
+            probe-repo --set $UUID --dest $UUID 2>&1 || true)
+          echo "$OUT"
+          echo "$OUT" | grep -q "restic not found"
+          echo "$OUT" | grep -q "/usr/bin/restic"
+          echo "$OUT" | grep -q "$RESTIC_STATION_DATA_DIR/config.json"
+          # "open Restic Station" is impossible advice on a headless host.
+          ! echo "$OUT" | grep -q "open Restic Station"
+
+          # 2. A restic on PATH is discovered and used, with no configuration
+          #    at all — the "headless host" acceptance criterion. Note it is
+          #    only "found" because it *runs*: an inert +x file would not do.
+          FAKEBIN=$(mktemp -d)
+          cat > "$FAKEBIN/restic" <<'SH'
+          #!/bin/sh
+          if [ "$1" = "version" ]; then
+            echo '{"version":"0.18.1","go_version":"go1.22.0","go_os":"linux","go_arch":"arm64"}'
+            exit 0
+          fi
+          exit 1
+          SH
+          chmod +x "$FAKEBIN/restic"
+          OUT=$(env PATH="$FAKEBIN" .build/debug/restic-station-helper \
+            probe-repo --set $UUID --dest $UUID 2>&1 || true)
+          echo "$OUT"
+          echo "$OUT" | grep -q "discovered $FAKEBIN/restic"
+          # Got past restic resolution and on to the real (absent) set.
+          echo "$OUT" | grep -q "no backup set with id"
+
+          # 3. tick still exits 0 on a host with nothing configured.
+          #    (tick resolves restic through the same `resolveResticPath`
+          #    that step 2 exercises, via `makeTolerant`.)
+          env PATH="$FAKEBIN" .build/debug/restic-station-helper tick
+
   macos:
     runs-on: macos-15
     steps:
--- a/App/Sources/Support/ResticStatusMapping.swift
+++ b/App/Sources/Support/ResticStatusMapping.swift
@@ -1,0 +1,45 @@
+import Foundation
+import ResticStationCore
+
+/// `ResticDiscovery`, `ResticProbe` and `ResticDiscoveryResult` live in Core
+/// (`Core/Sources/ResticStationCore/Restic/ResticDiscovery.swift`) since T25 —
+/// the helper needs them too, and `App/` is never compiled on Linux.
+///
+/// What stays here is only the *UI vocabulary*: the translation from a probe
+/// outcome into `ResticStatus`, the enum the Settings chip and the menu-bar
+/// health derivation speak. Core has no business knowing about it, and the
+/// mapping is the one part of the old file that was genuinely app-side.
+extension ResticProbe {
+    /// Maps a probe onto the status vocabulary the Settings chip and the
+    /// menu-bar health derivation already speak (`AppModel.resticStatus`).
+    var status: ResticStatus {
+        switch outcome {
+        case .ok(let version):
+            return .ok(path: path, version: version)
+        case .tooOld(let version):
+            return .tooOld(path: path, version: version, minimum: ResticDiscovery.minimumVersion)
+        case .unusable(let reason):
+            return .unavailable(path: path, reason: reason)
+        }
+    }
+}
+
+extension ResticDiscoveryResult {
+    /// The best thing we can say about this search, in the vocabulary of the
+    /// Settings chip: a working binary, else the first too-old one we found
+    /// (candidates are probed in preference order, so this is the binary the
+    /// user is most likely to think of as "their" restic), else the first
+    /// unusable candidate, else "nothing on this Mac".
+    var status: ResticStatus {
+        if let chosen {
+            return chosen.status
+        }
+        if let firstTooOld {
+            return firstTooOld.status
+        }
+        if let firstUnusable = rejected.first {
+            return firstUnusable.status
+        }
+        return .notConfigured
+    }
+}
--- a/App/Sources/ViewModels/AppModel.swift
+++ b/App/Sources/ViewModels/AppModel.swift
@@ -232,10 +232,11 @@ final class AppModel: ObservableObject {
             setHealths: setHealths,
             anyRunInFlight: !stateWatcher.currentRuns.isEmpty,
             // Only a *definite* denial is a warning: a missing
-            // `fda-check.json` means the probe has never run, which the
-            // Permissions pane reports as "unknown" (T18) rather than as a
-            // problem.
-            fullDiskAccessDenied: stateWatcher.fdaCheck.map { !$0.hasFullDiskAccess } ?? false,
+            // `fda-check.json` means the probe has never run (or does not
+            // apply on this platform), which the Permissions pane reports as
+            // "unknown" (T18) rather than as a problem. The rule lives in
+            // Core so the Linux build is held to it too (T25).
+            fullDiskAccessDenied: HealthDerivation.fullDiskAccessDenied(from: stateWatcher.fdaCheck),
             backgroundAgentEnabled: launchd.isEnabled
         )
     }
--- a/Core/Sources/ResticStationCore/Restic/ResticDiscovery.swift
+++ b/Core/Sources/ResticStationCore/Restic/ResticDiscovery.swift
@@ -1,42 +1,68 @@
 import Foundation
-import ResticStationCore
 
 /// Finds a usable `restic` binary (`docs/architecture.md` §restic discovery,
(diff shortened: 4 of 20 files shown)


bherila:
@codex review

Issue #31, opened 2026-08-05
T29 — Static Linux binary, release packaging, Linux integration test


bherila:
# T29 — Static Linux binary, release packaging, Linux integration test

**Size:** M · **Model:** Sonnet · **Depends on:** T21, T26, T27 · **Milestone:** M5

## Repository context (read first)
T21 gets `swift build` working on Linux; this task makes the result *distributable*. The target user is someone who wants to drop a binary on a NAS or a VPS and have it work — requiring them to install a Swift runtime first would defeat the purpose.

`scripts/integration-test.sh` already exists and exercises the helper end to end against a **real** restic binary, using `RESTIC_STATION_DATA_DIR` to redirect all state into a temp dir (see `docs/testing.md` for the script's contract). `.github/workflows/ci.yml` runs it on macOS. It is the natural basis for Linux end-to-end coverage — read it before writing anything new.

## Goal
A single self-contained `restic-station-helper` binary per architecture, published as a CI artifact with the systemd units and an install script, and proven by a real end-to-end restic run on Linux.

### 1. Static build
- Build with Apple's **Static Linux SDK** (musl), producing a binary with no libc, no Swift-runtime, and no ICU dependency. Cross-compiling from macOS works and is the fastest path for local release builds; CI may build in-container instead. Whichever you choose, document the exact toolchain and SDK version — Swift static-SDK invocations are version-sensitive and a reproducible command line is most of the value here.
- `--static-swift-stdlib` against glibc is the fallback if the static SDK proves impractical; it still leaves a glibc version floor, so note that floor in the docs if you land on it.
- Targets: `x86_64` and `aarch64`.
- **Verify staticness rather than assuming it**: `ldd` reports "not a dynamic executable", and the binary runs in a `scratch`/`alpine` container and on an old-glibc distro. Foundation on musl has historically had rough edges — if you hit one, record precisely what broke.

### 2. Release packaging (`scripts/package-linux.sh`)
Produce `restic-station-linux-<arch>.tar.gz` containing: the binary, `packaging/linux/systemd/` units from T26, an `install.sh` (installs to `~/.local/bin` by default or `/usr/local/bin` with `--system`, then points the user at `helper timer install`), the LICENSE, and a short README covering first-run setup. Emit a SHA256SUMS file. Keep the layout obvious enough to be unpacked and inspected by hand.

### 3. Linux integration test
[...]

PR #36, opened 2026-08-05
T23 — SecretStore abstraction + Linux file backend + secret CLI

### LINKED ISSUE #25: T23 — SecretStore abstraction + Linux file backend + secret CLI

bherila:
Abstracts secret storage behind a `SecretStore` protocol, keeps the macOS keychain backend behaviourally identical, adds a Linux file backend that needs no desktop session / keyring daemon / GPG agent, and adds terminal subcommands so secrets can be entered on a host with no GUI.

Rebased onto `0f48871` (T25), which is current `origin/main`.

### Core — `Core/Sources/ResticStationCore/Secrets/`
- **`SecretStore.swift`** — `public protocol SecretStore: Sendable` with the six existing operations (`setPassword`/`password`/`deletePassword`, `setSecretEnv`/`secretEnv`/`deleteSecretEnv`), `passwordCommand(destId:)` promoted from `static` to an instance requirement, and one addition, `passwordCommandEnvironment` (see below). `KeychainError` is gone, replaced by a backend-neutral `SecretStoreError` with `.itemNotFound` / `.backendFailed(String)` — one live error type, no deprecated alias. Shared `SecretAccount` (key naming) and `SecretEnvBlob` (JSON encoding) mean the two backends structurally cannot drift.
- **`KeychainSecretStore.swift`** — the old `KeychainClient`, renamed via `git mv`, gated `#if os(macOS)`. Zero behavioural change: same argv, same delete-then-add, same `-T /usr/bin/security`, same exit-44 mapping, same trailing-newline trim.
- **`FileSecretStore.swift`** — `<AppPaths.root>/secrets.json`.
  - Created **at** `0600` via `open(2)`'s mode and the directory **at** `0700` via `mkdir(2)`'s — never create-then-`chmod`. A `fchmod`/`chmod` immediately after creation pins the mode against the umask; since the file was *created* at `0600` and a umask can only clear bits, it is never wider than `0600` for an instant.
  - Every read re-verifies: `open(O_RDONLY|O_NOFOLLOW)` then `fstat` **on the opened descriptor** (closing both the symlink-swap and the TOCTOU holes), refusing a symlink, a non-regular file, or any group/other permission bit with a message naming the file and the exact `chmod` to run. `0400` is accepted — the rule is "no group/other bits", not "exactly `0600`".
  - Writes: unlink any stale temp file (so `O_CREAT|O_EXCL` really creates, since `open(2)` ignores `mode` for an existing file) → write → `fsync` → `rename(2)`, under `locks/secrets.lock` (the existing `FileLock`, 10 s bounded wait). Reads take no lock — `rename(2)` is atomic. A file whose `version` is newer than this build is refused rather than overwritten.
- **`SecretBackend.swift`** — `SecretBackend` (platform default; `RESTIC_STATION_SECRET_BACKEND=keychain|file` override, unrecognised value is a hard error, empty treated as unset like every other env override here) and `SecretStoreFactory`. `keychain` off macOS is a hard error naming the fix.

### Core — consumers
- `ResticRunner` and `BackupEngine` take `any SecretStore` (parameter renamed `keychain:` → `secrets:`).
- `ResticRunnerError.keychainUnavailable` → `.secretsUnavailable`. No deprecated alias: every call site is in this repo and was updated.
- macOS user-facing strings are preserved **verbatim** behind `#if os(macOS)`: `Reachability`'s `.offline(reason: "keychain locked")` (which `SetsBadges` matches on), `ResticRunnerError.userFacingMessage`'s "Unlock your login keychain…", and `BackupEngine`'s "the login keychain could not be read for destination …". Linux gets its own wording.

### Helper
- **`Commands/Secret.swift`** — `secret set`, `secret set-env`, `secret rm [--env]`, `secret list`, plus a hidden `print-password`. Values are read from **stdin**: echo-disabled `termios` prompt on a TTY (prompt goes to stderr, since stdout may be redirected), raw to EOF from a pipe with **exactly one** trailing newline stripped. `print-password` writes raw bytes with no trailing newline and is `shouldDisplay: false`.
- `HelperContext` selects the backend in one place; a new lightweight `SecretContext` backs the `secret` subcommands so entering a password works before restic is configured (`HelperContext.make()` exits 1 without a restic path). `print-password` loads no config at all — the password path should have as few moving parts as possible.
- Adapted to T25's now-`async` `HelperContext.make()`.

### App
- Constructs the store through `SecretStoreFactory` in all five places; `MountController` mirrors `ResticRunner`'s env assembly including `passwordCommandEnvironment`. `SetsBadges` learned the Linux offline reason.

### Docs
- `docs/keychain-and-fda.md` — retitled *Secret storage, Full Disk Access, and SMAppService*, backend table up front, and a new **§5 Linux: the `secrets.json` file backend** covering why a file rather than a keyring, the on-disk shape, an explicit **threat model with its out-of-scope list spelled out** (root, the invoking user, an unencrypted disk, memory scraping), the `0600` rules, concurrency, how the password reaches restic, why the file backend does *not* reproduce macOS's argv tradeoff, terminal usage, and troubleshooting. **Filename kept** (`keychain-and-fda.md`): it is referenced from a dozen source comments, task files and other docs, and #35 had just edited it; the H1 carries the accurate title and a note explains the choice.
- `docs/restic-cli.md` — the env table now reflects the per-backend password command, documents `passwordCommandEnvironment`, and records the empirically verified `RESTIC_PASSWORD_COMMAND` splitting rules (below).
- `docs/architecture.md`, `docs/testing.md` — updated for the protocol and the new test layout.

## The one non-obvious design decision

`ResticRunner` *replaces* restic's environment, and restic's password-command child inherits that replaced environment. On the file backend that child is **another copy of our own helper**, which re-resolves its store from its own environment — so it would look in the *default* data directory with the *platform-default* backend and fail to find a password this process had just read successfully. Found while running the end-to-end script: with `RESTIC_STATION_SECRET_BACKEND=file` on macOS, `restic init` (which inherited my shell) succeeded while `helper run-set` failed.

Fix: `SecretStore.passwordCommandEnvironment`, merged into the assembled environment just before `RESTIC_PASSWORD_COMMAND` (so `nonSecretEnv` still cannot hijack it). `FileSecretStore` returns `RESTIC_STATION_DATA_DIR` + `RESTIC_STATION_SECRET_BACKEND=file`; the protocol extension returns `[:]`, so `KeychainSecretStore` contributes nothing and **macOS's assembled environment is byte-identical to before**. Both values are non-secret by construction: a path and the word "file".

Related: the helper's own path comes from `/proc/self/exe` (Linux) or `_NSGetExecutablePath` + `realpath` (Darwin) — never `argv[0]` (attacker-controlled, and this string is executed by restic), and never `Bundle.main.executablePath`, which for a tool inside `Contents/MacOS/` names the **app**, not the tool. The app-bundle case is exercised in CI.

## `RESTIC_PASSWORD_COMMAND` splitting — verified, not assumed

The file backend's command names a path that can contain a space (`…/Restic Station.app/Contents/MacOS/restic-station-helper`), so I measured restic 0.18.1's own splitter rather than guessing, and recorded it in `docs/restic-cli.md`:

| Form | Result |
|---|---|
| `"…/with spaces/helper" print-password --dest <uuid>` | works; double-quoted args may contain spaces, `'`, `$`, `\`, and unquoted args may follow |
| `'…/with spaces/helper'` | works |
| `…/with spaces/helper` unquoted | `fork/exec …/with: no such file or directory` |
| `'…/it'\''s/helper'` (POSIX escape) | **fails** — no backslash escapes |
| a path containing `"` | `double-quoted string not terminated` — unrepresentable |

`quoteForRestic(_:)` therefore double-quotes only when the path contains a character outside `[A-Za-z0-9/._+=:,@-]`, so the common case emits the same bare shape the keychain backend does. Also confirmed: restic does **not** use `sh -c`, matching what the doc already claimed.

I did **not** switch Linux to a plain `RESTIC_PASSWORD` env var — the `_COMMAND` seam is kept on both platforms, as the issue directs.

### Automated (all green locally, all wired into CI)
- `swift build` + `swift test` (root, 22 tests) — includes new `SecretCommandTests` and the updated subcommand registration test.
- `swift test --package-path Core` — **369 tests, 56 suites**, up from 308.
- `./scripts/bootstrap.sh && xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO` — clean, no warnings.
- `scripts/integration-test.sh` (real keychain, real restic, real helper) — **ALL ASSERTIONS PASSED (27s)**. This is the direct evidence that macOS behaviour is unchanged.
- `RESTIC_STATION_KEYCHAIN_SMOKE=1 swift test --package-path Core --filter KeychainSmoke` — 2/2, real login keychain, prompt-free.

Test-layout changes:
- `ResticRunnerTests`, `ReachabilityTests` and `BackupEngineTests` now inject a `FakeSecretStore` instead of scripting `/usr/bin/security` through `FakeProcessRunner`. Their process scripts contain restic calls **and nothing else**, which both shortens them and makes them platform-independent.
- `KeychainSecretStoreTests` is the old `KeychainClientTests` with only the type name and the error-case name edited (`KeychainClient` → `KeychainSecretStore`, `KeychainError.securityCommandFailed` → `SecretStoreError.backendFailed`), plus `passwordCommand` now being called on an instance. Every assertion is unchanged.
- New `SecretStoreConformanceTests` runs the shared semantics (`.itemNotFound` on a missing password, `[:]` on a missing secret env, idempotent deletes, overwrite-not-append) against **both** backends; `FileSecretStoreTests` and `SecretBackendTests` are **not** gated to Linux. That is deliberate: forcing the file backend on macOS is the only way this repo can genuinely cover the Linux secrets path today.

### The no-secret-leak property — exactly how it was verified

Four independent layers, because "no secret leaks" is the claim most easily asserted and least easily earned:

1. **Structural.** On the Linux path a secret has exactly three legal locations: the bytes of `secrets.json`, the stdin of `secret set`/`set-env`, and the stdout of `print-password`. `secret list`'s renderer (`SecretListing.format`) is handed only ids, labels, set names and **counts** — it never receives a secret value, so it cannot print one. `SecretStoreError.backendFailed` may only carry a backend's own diagnostic text (a path, an errno, a JSON key path); `ResticRunnerError.secretsUnavailable` carries only a destination id.
2. **Unit assertions.** `FileSecretStoreTests.errorsAndPathsNeverCarrySecrets` stores a random secret, then asserts it appears in neither the password command, nor the file path, nor the lock path, nor the error raised when the mode is widened. `SecretCommandTests` asserts the `set-env` parser names the offending variable but never any value, and never echoes malformed input back. `ResticRunnerTests.fileBackendEnvironmentAssembly` asserts the password is in neither restic's argv nor its env (only the *command* is), while the secret-env blob is still injected as real variables.
3. **Process-level, against the real binary.** `scripts/secret-cli-test.sh` runs the whole flow with `RESTIC_STATION_SECRET_BACKEND=file` and a deliberately awkward password — `p@ss "word" with $spaces and a trailing space ` — asserting: `secret set` reads a **pipe**; `secrets.json` is `0600`; `print-password` returns the **exact bytes** (hex-dump comparison, not string comparison) with **no trailing newline**; `echo`-piped input round-trips (exactly one newline stripped); `secret list`'s actual stdout is grepped for the password, the access key id and the secret env value and contains **none of them**; a `0644` file is refused with a message that names the file and the `chmod` and does **not** echo the secret.
4. **Whole-data-directory scan after a real run.** After `helper run-set --kind backup` completes against a real restic repo, the script runs `grep -rlF "$PASSWORD" "$DATA_DIR"` and asserts the **only** matching file is `secrets.json` itself — covering `runs/<id>/log.txt`, `runs/index.jsonl`, `runs/<id>/metadata.json` (including `argvRedacted`), every `state/*.json`, and `config.json`. The same scan runs for the secret-env value.

Not covered by these: the macOS keychain backend still exposes `security -w <value>` in argv. That is the pre-existing, documented, accepted tradeoff — untouched, and explicitly **not** reproduced by the file backend.

### The restic-authenticates acceptance criterion

Cannot be run on a Linux host from this environment (no Linux toolchain, no container runtime). Done on macOS with `RESTIC_STATION_SECRET_BACKEND=file` against a real local restic repo, using the helper **inside the built app bundle** — whose path contains a space, so this also exercises the quoting rule and the `_NSGetExecutablePath` resolution:

```
=== 7. real restic authenticates via RESTIC_PASSWORD_COMMAND ===
restic 0.18.1 compiled with go1.25.1 on darwin/arm64
RESTIC_PASSWORD_COMMAND=["/Users/bwh/.../dd/Build/Products/Debug/Restic Station.app/Contents/MacOS/restic-station-helper" print-password --dest 00000000-0000-4000-8000-0000000000c1]
ok: restic init opened the repo using the helper's print-password
backup success (1 run)
ok: helper run-set produced a snapshot through FileSecretStore

=== 8. the password appears in no log, run record, state file or config ===
ok: the only file containing the password is secrets.json itself
ok: no secret-env value outside secrets.json either
```

Full transcript of a run against the SwiftPM-built helper:

```
=== 1. secret set reads the password from a pipe (never argv) ===
stored a password for "Primary"
ok: secret set stored a password

=== 2. file is 0600, directory is 0700 ===
ok: secrets.json=600 data dir=700
ok: keyed on the lowercased destination uuid, mirroring the keychain accounts

=== 3. print-password returns the exact bytes, with no trailing newline ===
ok: byte-identical, 46 bytes, last byte 0x20
ok: exactly one trailing newline is stripped from piped input

=== 4. secret set-env, and secret list never prints a value ===
stored 2 secret environment variable(s) for "Primary": AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
00000000-0000-4000-8000-0000000000c1  "Primary" (set "Secret CLI Set")  password, secret-env (2 variable(s))
ok: secret list names destinations only — no password, no key, no value
ok: set-env rejects a non-object with exit 1

=== 5. an unknown destination exits 1 ===
ok: unknown destination → exit 1

=== 6. a widened mode is refused with an actionable message ===
could not read the password for destination …: refusing to read …/data/secrets.json: it is
group- or world-accessible (mode 0644). Fix it with: chmod 600 …/data/secrets.json
ok: 0644 refused, naming the file and the fix, without echoing the secret

=== 9. secret rm is idempotent ===
"Primary": no password is stored any more
"Primary": no password is stored any more
"Primary": no secret environment is stored any more
"Primary": no secret environment is stored any more
ok: two removes of each kind, both exit 0
ok: the password really is gone
ok: secret list reports an empty store

=== ALL SECRET CLI ASSERTIONS PASSED ===
```

`scripts/secret-cli-test.sh` is wired into **both** CI jobs: the Linux job runs it against the SwiftPM binary (skipping the restic half — restic is not in the `swift:6.1` container), and the macOS job runs it against the app-bundle helper *with* restic, so the transcript above is reproduced on every CI run.

## Acceptance criteria

| Criterion | Status |
|---|---|
| macOS behaviour provably unchanged; existing keychain tests pass with only the type name edited | **Verified locally** — the `KeychainSecretStoreTests` diff is type/error names only; `integration-test.sh` (real keychain + real restic + real helper) fully green; keychain smoke test green |
| `swift test` green on macOS and Linux; keychain backend properly `#if os(macOS)`-scoped | macOS **verified locally** (369 Core + 22 root). Linux **verified in CI** (see checks). Note: T21 added no gates to un-gate — Core already built green on Linux; the new scoping is on `KeychainSecretStore` and its tests only |
| restic authenticates via `RESTIC_PASSWORD_COMMAND` against `FileSecretStore`, with a transcript | **Verified on macOS** with `RESTIC_STATION_SECRET_BACKEND=file` against a real repo (transcript above), and in macOS CI. **Linux-host verification deferred to #31 (T29)** — no Linux toolchain or container runtime is available here |
| No secret in argv, logs, run records, or errors on the Linux path | **Verified** — four layers, detailed above; strongest is the post-run `grep -rlF` of the whole data directory |
| `docs/keychain-and-fda.md` gains a Linux section (file backend, threat model, `0600` enforcement, contrast with the macOS ACL strategy); retitle if the filename misleads | **Done** — §5 added, doc retitled. Filename deliberately kept, with the reason stated in the doc |

### Also deferred
- Linux-host execution of anything beyond compile + unit + CLI tests (the point of #31 / T29). Everything Linux-specific here is verified by (a) the file backend running under macOS CI and locally, and (b) the Linux CI job running the unit and CLI suites.
- `FileSecretStore` under a **non-root uid on Linux with a hostile sibling user** — the threat model is documented, but only the mode enforcement is mechanically tested.
- Interactive TTY behaviour (`secret set` with echo disabled): the pipe path is asserted end to end; the `isatty` branch is exercised by hand only, since CI has no TTY.

## Scope discipline
No change to the config schema (#26), restic discovery (#27), or `AppPaths.swift` (owned elsewhere — read only; `FileSecretStore` composes its lock path from `paths.locksDir` and notes in a comment that it belongs in `AppPaths` later).

Closes #25

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,13 +21,29 @@ jobs:
           OUT=$(.build/debug/restic-station-helper --help)
           echo "$OUT"
           # `timer` is Linux-only (T26); the macos job asserts its absence.
-          for sub in tick run-set init-secondary restore probe-repo unlock fda-check timer version; do
+          for sub in tick run-set init-secondary restore probe-repo unlock fda-check secret timer version; do
             echo "$OUT" | grep -qE "^  $sub( |$)" || { echo "missing subcommand: $sub"; exit 1; }
           done
+          # print-password is registered but deliberately hidden: it is the
+          # machinery behind RESTIC_PASSWORD_COMMAND, not a user action.
+          if echo "$OUT" | grep -qE "^  print-password( |$)"; then
+            echo "print-password must not appear in --help"
+            exit 1
+          fi
+          # …hidden, but reachable.
+          .build/debug/restic-station-helper print-password --help >/dev/null
+          .build/debug/restic-station-helper secret --help
       # SwiftPM does not run a path dependency's test targets, so Core's
       # suite still needs its own invocation.
       - name: Run Core unit tests
         run: swift test --package-path Core
+      # The Linux secret backend, end to end against the real binary. restic
+      # is not installed in this container, so the script skips its restic
+      # half (T29 owns the Linux restic matrix) and still asserts the file
+      # modes, the exact print-password bytes, and that `secret list` prints
+      # no secret value.
+      - name: Secret CLI + FileSecretStore end to end
+        run: scripts/secret-cli-test.sh
 
       # T25 acceptance criteria that only a real Linux run can show.
       - name: fda-check is a no-op on Linux
@@ -205,6 +221,12 @@ jobs:
         run: swift test --package-path Core
       - name: Run integration test
         run: scripts/integration-test.sh
+      # Forces the Linux file backend on macOS, where restic IS installed —
+      # so this run also proves the acceptance criterion "restic
+      # authenticates via RESTIC_PASSWORD_COMMAND against FileSecretStore",
+      # against the helper inside the real app bundle.
+      - name: Secret CLI + FileSecretStore end to end (file backend on macOS)
+        run: scripts/secret-cli-test.sh "dd/Build/Products/Debug/Restic Station.app/Contents/MacOS/restic-station-helper"
       - name: Ad-hoc sign and package app artifact
         run: |
           APP="dd/Build/Products/Debug/Restic Station.app"
--- a/App/Sources/ViewModels/AppModel+Maintenance.swift
+++ b/App/Sources/ViewModels/AppModel+Maintenance.swift
@@ -40,7 +40,7 @@ enum MaintenanceLookup {
         return ResticRunner(
             resticPath: resticPath,
             paths: model.paths,
-            secrets: try SecretStoreFactory.make(paths: model.paths, runner: processRunner),
+            secrets: try model.makeSecretStore(),
             runner: processRunner
         )
     }
--- a/App/Sources/ViewModels/AppModel+Maintenance.swift
+++ b/App/Sources/ViewModels/AppModel+Maintenance.swift
@@ -40,7 +40,7 @@ enum MaintenanceLookup {
         return ResticRunner(
             resticPath: resticPath,
             paths: model.paths,
-            keychain: KeychainClient(runner: processRunner),
+            secrets: try SecretStoreFactory.make(paths: model.paths, runner: processRunner),
             runner: processRunner
         )
     }
--- a/App/Sources/ViewModels/AppModel+Restore.swift
+++ b/App/Sources/ViewModels/AppModel+Restore.swift
@@ -72,10 +72,11 @@ extension AppModel {
     func makeBrowsingRunner() -> ResticRunner? {
         guard let path = config.resticPath, !path.isEmpty else { return nil }
         let processRunner = DefaultProcessRunner()
+        guard let secrets = try? SecretStoreFactory.make(paths: paths, runner: processRunner) else { return nil }
         return ResticRunner(
             resticPath: path,
             paths: paths,
-            keychain: KeychainClient(runner: processRunner),
+            secrets: secrets,
             runner: processRunner
         )
     }
--- a/App/Sources/ViewModels/AppModel+Restore.swift
+++ b/App/Sources/ViewModels/AppModel+Restore.swift
@@ -69,14 +69,20 @@ extension AppModel {
     ///
     /// Returns `nil` when no restic binary is configured yet; the caller
     /// renders that as an explained empty state rather than an error.
-    func makeBrowsingRunner() -> ResticRunner? {
+    ///
+    /// **Throws** when secret storage itself is misconfigured (a mistyped
+    /// `RESTIC_STATION_SECRET_BACKEND`). That is deliberately not folded into
+    /// the `nil` case: "restic is not set up yet" is an empty state, while
+    /// "your secret storage setting is wrong" is a real error every other
+    /// flow reports — swallowing it here would leave the Restore pane showing
+    /// a bare "No snapshots" with no hint of the cause.
+    func makeBrowsingRunner() throws -> ResticRunner? {
         guard let path = config.resticPath, !path.isEmpty else { return nil }
         let processRunner = DefaultProcessRunner()
-        guard let secrets = try? SecretStoreFactory.make(paths: paths, runner: processRunner) else { return nil }
         return ResticRunner(
             resticPath: path,
             paths: paths,
-            secrets: secrets,
+            secrets: try makeSecretStore(),
             runner: processRunner
         )
     }
@@ -173,11 +179,19 @@ final class RestoreBrowser: ObservableObject {
 
     /// Points the browser at a repository. Re-configuring with the same
     /// repository is a no-op, so it is safe to call from `onChange`.
-    func configure(repository: RestoreRepository?, runner: ResticRunner?) {
+    ///
+    /// - Parameter unavailableReason: why there is no runner, when the
+    ///   reason is a *misconfiguration* rather than "restic is not set up
+    ///   yet". Surfaced through `snapshotsError`, so the pane explains
+    ///   itself instead of showing a bare "No snapshots".
+    func configure(repository: RestoreRepository?, runner: ResticRunner?, unavailableReason: String? = nil) {
         guard repository?.id != self.repository?.id else {
             // Same repository, but the restic path (and therefore the
             // runner) may have changed in Settings.
             self.runner = runner
+            if let unavailableReason {
+                snapshotsError = unavailableReason
+            }
             return
         }
         snapshotsTask?.cancel()
@@ -185,7 +199,9 @@ final class RestoreBrowser: ObservableObject {
         self.repository = repository
         self.runner = runner
         snapshots = []
-        snapshotsError = nil
+        // Set after the reset, not before: `loadSnapshots` bails out early
+        // without a runner, so this is the only thing that will be showing.
+        snapshotsError = unavailableReason
         isLoadingSnapshots = false
         searchResults = []
         searchState = .idle
--- a/App/Sources/ViewModels/AppModel+Sets.swift
+++ b/App/Sources/ViewModels/AppModel+Sets.swift
@@ -20,11 +20,32 @@ extension AppModel {
     /// keychain by default, or `secrets.json` when
     /// `RESTIC_STATION_SECRET_BACKEND=file`. Cheap to build.
     ///
-    /// `nil` only when the backend override is unrecognised, which the
-    /// callers surface as "the password could not be saved" rather than
-    /// silently writing to the wrong store.
+    /// **The helper path is not this process.** With the file backend the
+    /// store bakes an executable into `RESTIC_PASSWORD_COMMAND`, and restic
+    /// runs it as a child to read the password. Naming the app binary there
+    /// would hand restic a SwiftUI app that cannot print a password and
+    /// would try to open a UI, so every app-side store names the *embedded
+    /// helper* — the same binary `HelperInvoker` and the LaunchAgent use.
+    /// The factory requires this explicitly, so a new app-side call site
+    /// cannot forget it.
+    ///
+    /// Throws only when the `RESTIC_STATION_SECRET_BACKEND` override is
+    /// unrecognised — deliberately a hard error rather than a silent
+    /// fallback to the wrong store.
+    func makeSecretStore() throws -> any SecretStore {
+        try SecretStoreFactory.make(
+            paths: paths,
+            runner: DefaultProcessRunner(),
+            helperExecutablePath: HelperInvoker.helperURL.path
+        )
+    }
+
+    /// `makeSecretStore()` for the call sites that can only degrade (reading
+    /// back a password to pre-fill an editor, deleting a destination's
+    /// secrets). Every site that can *report* the failure calls the throwing
+    /// form instead.
     var secrets: (any SecretStore)? {
-        try? SecretStoreFactory.make(paths: paths, runner: DefaultProcessRunner())
+        try? makeSecretStore()
     }
 
     // MARK: - Sets
@@ -111,7 +132,7 @@ extension AppModel {
         password: String?,
         secretEnv: [String: String]?
     ) async throws {
-        let store = try SecretStoreFactory.make(paths: paths, runner: DefaultProcessRunner())
+        let store = try makeSecretStore()
         if let password, !password.isEmpty {
             try await store.setPassword(password, destId: destId)
         }
--- a/App/Sources/ViewModels/AppModel+Sets.swift
+++ b/App/Sources/ViewModels/AppModel+Sets.swift
@@ -15,11 +15,16 @@ extension AppModel {
 
     // MARK: - Collaborators
 
-    /// A keychain client bound to the same `security(1)` mechanics the helper
-    /// uses (`docs/keychain-and-fda.md` §1). Cheap to build — it is a struct
-    /// wrapping a process runner.
-    var keychain: KeychainClient {
-        KeychainClient(runner: DefaultProcessRunner())
+    /// The secret store this app writes destination passwords into — the
+    /// same one the helper reads (`docs/keychain-and-fda.md`): the login
+    /// keychain by default, or `secrets.json` when
+    /// `RESTIC_STATION_SECRET_BACKEND=file`. Cheap to build.
+    ///
+    /// `nil` only when the backend override is unrecognised, which the
+    /// callers surface as "the password could not be saved" rather than
+    /// silently writing to the wrong store.
+    var secrets: (any SecretStore)? {
+        try? SecretStoreFactory.make(paths: paths, runner: DefaultProcessRunner())
     }
 
     // MARK: - Sets
@@ -106,15 +111,15 @@ extension AppModel {
         password: String?,
         secretEnv: [String: String]?
     ) async throws {
-        let keychain = self.keychain
+        let store = try SecretStoreFactory.make(paths: paths, runner: DefaultProcessRunner())
         if let password, !password.isEmpty {
-            try await keychain.setPassword(password, destId: destId)
+            try await store.setPassword(password, destId: destId)
         }
         if let secretEnv {
             if secretEnv.isEmpty {
-                try await keychain.deleteSecretEnv(destId: destId)
+                try await store.deleteSecretEnv(destId: destId)
             } else {
-                try await keychain.setSecretEnv(secretEnv, destId: destId)
+                try await store.setSecretEnv(secretEnv, destId: destId)
             }
         }
     }
@@ -123,18 +128,18 @@ extension AppModel {
     /// password is not an error here (a destination can exist before its
     /// password was ever stored) — the caller shows "leave blank to keep".
     func loadDestinationSecrets(destId: UUID) async -> (password: String?, secretEnv: [String: String]) {
-        let keychain = self.keychain
-        let password = try? await keychain.password(destId: destId)
-        let env = (try? await keychain.secretEnv(destId: destId)) ?? [:]
+        guard let store = self.secrets else { return (nil, [:]) }
+        let password = try? await store.password(destId: destId)
+        let env = (try? await store.secretEnv(destId: destId)) ?? [:]
         return (password, env)
     }
 
-    /// Both keychain items for a destination, as promised by the remove
-    /// confirmation copy. Deletes are idempotent in `KeychainClient`.
+    /// Both stored secrets for a destination, as promised by the remove
+    /// confirmation copy. Deletes are idempotent in every `SecretStore`.
     func deleteDestinationSecrets(destId: UUID) async {
-        let keychain = self.keychain
-        try? await keychain.deletePassword(destId: destId)
-        try? await keychain.deleteSecretEnv(destId: destId)
+        guard let store = self.secrets else { return }
+        try? await store.deletePassword(destId: destId)
+        try? await store.deleteSecretEnv(destId: destId)
     }
 
     // MARK: - Probe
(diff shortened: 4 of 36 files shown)


bherila:
@codex review

bherila:
All three review findings fixed in `b1018d2`. Rebased onto `12f9012` (T26) first — the `HelperMain` / `ci.yml` / subcommand-test conflicts are resolved so `secret` and `print-password` sit alongside the Linux-only `timer`.

### 1 (P2) — file backend captured the wrong executable

Agreed, and reachable today. `FileSecretStore.init` and `SecretStoreFactory.make` now **require** `helperExecutablePath` — no default at all, so the decision is a compile error to skip rather than a runtime hang in a flow nobody exercises. The helper passes `FileSecretStore.currentExecutablePath()` (the one process for which "this executable" is right); the app passes `HelperInvoker.helperURL.path` through a single new `AppModel.makeSecretStore()`, so `makeBrowsingRunner`, `MountController`, `AppModel+Maintenance`, `storeDestinationSecrets` and primary `init` cannot diverge.

Regression test — `SecretBackendTests.factoryUsesTheGivenHelperPath`. It asserts the exact command string *and* that the command does not contain `FileSecretStore.currentExecutablePath()`; in a test process that is the test runner, standing in for "any wrong binary". That assertion fails if the default is ever reintroduced.

### 2 (P2) — messaging branched on OS, not backend

Agreed. `SecretStore` gained `var backend: SecretBackend`, and all wording now lives on `SecretBackend`: `displayName`, `unavailableSummary`, `unavailableAdvice`, `unavailableProbeReason`. `BackupEngine` and `Reachability` take it from the store they hold. `ResticRunnerError.userFacingMessage` is the one place with no store to ask — it is a property on an error value — so it uses `SecretBackend.configured`, which reads the same environment every store is built from; in production the two always agree, and it deliberately falls back to the platform default rather than throwing (construction already treats a bad value as fatal). Documented at the call site and in `keychain-and-fda.md`.

The `.keychain` strings are byte-for-byte the pre-T23 ones, including `"keychain locked"`, which `SetsBadges.offlineOrError` matches to classify the failure as environmental rather than as a repository error.

Regression tests, all asserting **both** backends rather than just the host's:
- `SecretBackendTests.wordingFollowsTheBackend` — pins the keychain strings verbatim, and asserts the file wording never contains "keychain".
- `ReachabilityTests.secretPreflightReasonFollowsTheBackend` — loops `SecretBackend.allCases`.
- `BackupEngineTests.rowNineReasonFollowsTheBackend` — loops `allCases`; the `.retryable` reason is what a user reads.
- `ResticErrorTests.secretsUnavailableWordingFollowsTheBackend`.

### 3 (P3) — `try?` swallowed the deliberate hard error

Agreed. `makeBrowsingRunner()` now `throws`; `RestoreView.configureBrowser()` catches and passes the reason to `RestoreBrowser.configure(…, unavailableReason:)`, which surfaces it through the existing `snapshotsError` state (the pane already has a proper "Snapshots could not be read" view with a Try Again button). "restic is not configured yet" stays the `nil` / empty-state case — the two really are different, which is why they are no longer conflated.

### Verification re-run in full

- `swift build` + `swift test` (root, 22) — green.
- `swift test --package-path Core` — **375 tests**, up from 369; the 6 new ones are the regression tests above.
- `./scripts/bootstrap.sh && xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO` — clean, no warnings.
- `scripts/secret-cli-test.sh` — green against the SwiftPM helper **and** against the app-bundle helper (path with a space), including the real-restic half and the whole-data-directory leak scan.
- `scripts/integration-test.sh` — ALL ASSERTIONS PASSED (26s). This is the macOS-unchanged guarantee.
- `RESTIC_STATION_KEYCHAIN_SMOKE=1 … --filter KeychainSmoke` — 2/2 against the real login keychain.
- CI green on both jobs; the macOS job again shows `ok: helper run-set produced a snapshot through FileSecretStore` and `ok: the only file containing the password is secrets.json itself`.

No disagreements with any of the three findings. Not merging — held for the Fable safety review.

PR #37, opened 2026-08-05
T24 — Per-machine config scoping (schema v2) + migration

### LINKED ISSUE #26: T24 — Per-machine config scoping (schema v2) + migration

bherila:
One `config.json` now describes a whole fleet. Each host additionally keeps a small, never-shared `machine.json`; `config.json` gains optional `machines` override maps keyed by that host's `machineId`.

## What changed

**New — `Core/.../Config/MachineConfig.swift`**
- `MachineConfig` (`version`, `machineId`, `resticPath?`) at `<AppPaths.root>/machine.json`, versioned independently of `AppConfig`.
- `MachineStore` — same conventions as `ConfigStore` (no caching, `.sortedKeys` + `.prettyPrinted`, temp file + `rename(2)`), except `load()` auto-creates the file. `RESTIC_STATION_MACHINE_ID` overrides the id for the process without rewriting the file; the environment is injected, not read at point of use, so tests are hermetic.
- `MachineIdentity.slugify/generate/isValid` — lowercase, non-`[a-z0-9-]` → `-`, collapse, trim; a lowercased UUID when the hostname is unusable (still a valid slug).
- An invalid `machineId` on disk is a **hard error**, not a silent regeneration: an id that matches no `machines` key resolves to "no overrides", which is exactly the silent-wrong-backup failure this design exists to prevent.

**Schema v2 — `Config/Models.swift`**
- `AppConfig.currentVersion` → 2. `AppConfig.resticPath` kept as a documented, deprecated fallback.
- `BackupSet.machines: [String: BackupSetMachineOverride]?` (`enabled?`, `sources?`, `schedule?`) and `Destination.machines: [String: DestinationMachineOverride]?` (`enabled?`, `repoURL?`, `nonSecretEnv?`).
- `machines` uses `encodeIfPresent` (the second deliberate exception after `onboardingCompleted`), so a config with no overrides is byte-identical apart from its version number rather than gaining `"machines": null` on every set and destination. Override fields are likewise sparse — `"sources": null` inside a patch reads like the opposite of what it means.

**New — `Config/Resolution.swift`**
- `ResolvedConfig { machineId, config: AppConfig, omissions: [ResolvedOmission] }`, plus `AppConfig.resolved(for machineId:)`, `resolved(for machine: MachineConfig)`, and `referencedMachineIds`.

**Validation** (invariants 6 and 7, in `Models.swift` — this repo has no `Config/Validate.swift`; validation lives next to the models)
- 6: `machines` keys must be valid slugs; override `sources` entries must be absolute; override `schedule` must be in range. An **empty** override `sources` array is deliberately legal where an empty top-level one is not — it is how a machine says "nothing to back up here", and resolution drops the set with a recorded reason rather than running a source-less backup.
- 7: per machine, if a set runs there, exactly one of its destinations must be an enabled primary. Deliberately stated **before** resolution drops anything, so "every destination disabled" cannot quietly become "set skipped".
- An unknown-machine key is accepted (warning material for T27's `config validate`, not an error).

**Migration** — `ConfigStore.load()` (details below).

**`ConfigDiff`** — `machines` added to the set projection; destination-level overrides ride along inside `Destination`. Treated as schedule-relevant even for another machine, erring the way the rest of that type errs (an extra tick that finds nothing due).

**Helper** — `HelperContext.loadResolvedConfig` is the single resolution point; `HelperContext.config` is now the resolved config and `context.resolved` carries the omissions. `tick` prints one line per omission. `run-set --set <id>` for a set that exists in the shared config but not on this machine explains *why* instead of "no backup set with id …".

**T25 seam** — both `TODO(#26)` markers in `Helper/Sources/ResticPathResolution.swift` are filled: `resolveResticPath` now takes a `ResolvedConfig` (steps 1+2 arrive already collapsed, in order, by `AppConfig.resolved(for: MachineConfig)` — and a config that skipped machine resolution can no longer be passed by mistake), and the Linux not-found message points at `machine.json`.

**App** — `AppModel` keeps `config` **raw** (round-trip safety: editors read and save it, `machines` keys it does not edit are never dropped) and adds `resolvedConfig` + `machine`. Everything *derived* — health, staleness, the restic chip — reads the resolved one. `AppModel.resticPath` is the single accessor; discovery now persists to `machine.json` via a new `updateMachine(_:)`. Per-machine editing UI is out of scope — see the deferred item below.

**Docs** — `docs/data-model.md` (config/machine split, v2 types, encoding conventions, invariants 6–7, §machine.json, §Per-machine scoping with the resolution algorithm and both worked examples, rewritten §Versioning & migration); `docs/architecture.md` (AppPaths table + the rsync caveat); `docs/restic-cli.md`; `docs/testing.md`.

## Resolution algorithm

For each set, in config order:
1. Start from the top-level values.
2. Apply `set.machines[machineId]` if present — **replacing** each field it specifies (never merging).
3. Drop the set if effective `enabled == false`; its destinations are not even evaluated.
4. For each destination in order, apply its override the same way; drop it if effective `enabled == false`.
5. Drop the set if it ends up with zero enabled destinations, or zero sources while still enabled.

Every drop records a `ResolvedOmission` (`disabledForMachine` | `noEnabledDestinations` | `noSources`). Absent `machines`, or no entry for this machine, means inherit and run. Every `machines` map is stripped from the resolved value, so it cannot be resolved twice or saved back over the shared file's overrides. `resolved(for: MachineConfig)` additionally collapses `machine.resticPath ?? config.resticPath` into `config.resticPath`.

The function contains no platform branch, no filesystem access and no environment lookup: it depends only on `machineId`.

## Migration behaviour

`ConfigStore.load()` of a `version: 1` file, in this order:
1. Adds **no** `machines` keys (absence already means "everywhere").
2. Moves `resticPath` into `machine.json` if that file has none, and clears the deprecated field **only once the write succeeded**. If `machine.json` already has a path, that one wins.
3. Copies the untouched v1 bytes to `config.v1.backup.json` with `O_EXCL` — never overwriting an existing backup (including one the user put there).
4. Writes the v2 config **only if that backup exists**.

Idempotent (a second load sees v2 and does nothing). Persistence is best-effort so an unwritable data directory cannot stop backups from running; the **ordering** is not — the v1 file is never overwritten without a backup. An invalid v1 config throws at its own version and produces neither a backup nor a rewrite.

Net effect on an existing install: `"version": 1` → `2`, `"resticPath"` → `null`, everything the engine acts on unchanged.

## How verified

Locally, on macOS (Swift 6.3.3):
- `swift build` + `swift test` (root package) — green, 10 tests.
- `swift test --package-path Core` — green, **410 tests in 64 suites**.
- `./scripts/bootstrap.sh && xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO -quiet` — green.
- `./scripts/integration-test.sh` — **all assertions passed (50s)**, against real restic 0.18.1 and the real keychain, including the new `assert_migration` step (generated `machineId` was `bens-laptop-local`).

Two real bugs were caught by the new tests before commit: the environment override was read from `ProcessInfo` instead of the injected snapshot, and invariant 7 originally checked *after* resolution, which let "primary disabled, and it was the only destination" fall through as a silent set-drop rather than an error.

Coverage added: resolution table (no `machines` key; entry for another machine only; disabled set; disabled destination; `sources`/`repoURL`/`schedule`/`nonSecretEnv` replacement; zero enabled destinations; empty override sources; order preservation; overrides stripped); cross-platform invariance (expected JSON as source literals, run on both platforms); migration (backup verbatim, idempotence, pre-existing backup untouched, invalid v1, no-restic-path, machine.json auto-creation, v2 no-op, newer-version refusal); `MachineStore` (slug table, auto-create, env override, invalid/newer/corrupt rejection, atomic save, documented on-disk shape); validation (invariants 6 and 7, unknown-machine key accepted); `ConfigDiff` per-machine changes; new v1/v2 fixtures.

## Acceptance criteria

- [x] `swift test` green on macOS — **verified locally**; on Linux — **verified in CI** (the Core suite runs in both jobs, so the cross-platform invariant test executes on both).
- [x] Existing v1 configs load, migrate and behave identically on the authoring machine — **verified locally** three ways: `Fixtures/config-v1.json` (a realistic pre-change config: two sets, three destinations, Homebrew restic path) round-trips to identical `sets` after migration + resolution; `resolved(for:)` is the identity on it for any `machineId`; and the integration test migrates a real v1 config mid-run and keeps backing up.
- [x] `BackupEngine` and `ScheduleMath` unchanged — **verified**: zero lines changed in either file; they consume `resolved.config`.
- [x] `docs/data-model.md` documents v2, both override types, the resolution algorithm, `machine.json`, and the migration, with both worked examples (Linux-as-source, Linux-as-mirror-only).
- [x] The macOS app still builds and round-trips `machines` keys it does not edit — **verified locally** (app builds; `AppModel.config` stays raw, and `v2FixtureRoundTrips` pins that no key is gained or lost).
- [ ] **Deferred:** per-machine editing UI. Explicitly out of scope per the issue; a follow-up issue should be filed. `SetsCopy` maps the three new `ConfigError` cases to messages that name `config.json` and the fix, so a hand-edited override that fails validation is still explained in the UI rather than shown as a raw error.
- [ ] **Deferred:** "say so in the Linux install docs" — there is no Linux install doc in the repo yet (the README's install section is macOS-only). The never-copy rule is stated in `MachineConfig`'s doc comment, in `docs/data-model.md` (twice: the file-split table and §machine.json), and in `docs/architecture.md`'s rsync note. It should be repeated in the Linux install doc when that task lands.

## Notes for review

- `AppPaths` gained two additive, root-relative computed properties (`machineFile`, `configV1BackupFile`) rather than deriving those paths elsewhere — the type's own doc comment says never to hard-code a path outside it. No existing member changed.
- `resolveResticPath`'s parameter changed from `AppConfig` to `ResolvedConfig`, which touched four T25 tests; two new tests cover machine-over-config precedence and the empty-machine-path fallback.
- Rebased onto `0f48871` (T25); `HelperContext.make()`/`makeTolerant` are `async` here and all call sites carry `await`.

Closes #26


bherila:
@codex review

bherila:
## Addressed all seven review findings (rebased onto `12f9012`, T26)

### Four resolution-gap findings — one root cause, one fix

Resolution reached the scheduler path but not the read/query/utility paths. `ResolvedConfig` now carries a `Scope`, and there are two explicitly named accessors sharing one implementation:

| | `AppConfig.resolved(for:)` → `.scheduling` | `AppConfig.addressable(for:)` → `.addressable` |
|---|---|---|
| Overrides applied | yes | yes |
| `enabled: false` drops things | **yes** | **no** |
| `omissions` | populated | always empty |
| Used by | `tick`, `run-set` (backup/check/prune), health/staleness | `restore`, `probe-repo`, `unlock`, `init-secondary`, restore browser, maintenance sizes + `forget --dry-run`, "Initialize repository" |

Because both views run the same override code, they can never disagree about what a repository *is* — only about which ones this machine backs up.

- **`HelperContext.swift:95`** — `HelperContext` now exposes `scheduled` and `addressable` and **no bare `config`**, so every subcommand has to state which it means. The four repository utilities take `addressable`; `run-set` and `tick` take `scheduled`. `BackupEngine` is constructed with the addressable view: its only use of a config is `locate(destId:)`, which maps a destination back to its owning set for the *set lock*, and a restore into a set this machine does not back up must still be serialized correctly. What gets backed up is decided by the `BackupSet` handed to `runSet`, which comes from `scheduled`.
- **`AppModel+Restore.swift:73`** — `restoreRepositories` builds from `addressableConfig`.
- **`AppModel+Maintenance.swift:33`** — `MaintenanceLookup.set` and `resolvedSet(in:)` resolve through `addressableConfig`, so sizes and the retention preview describe the repository the confirm action prunes.
- **`AppModel+Sets.swift:239`** — `initializeRepository` looks the destination up in `addressableConfig`.

Audited beyond the four named: the Restore and Maintenance **set pickers** also read raw `config.sets`. Behaviourally identical today (the addressable view never drops a set), but changed anyway so the picker and the lookup can never drift. Deliberately left on raw `config`: the set-list/editor screens (they edit the shared file, and must keep round-tripping `machines` keys) and run-history labels (raw is a superset, which is what you want for labelling past runs). New `ResolvedConfig.set(id:)` / `destination(id:)` / `destinations` are now the sanctioned way to find a repository, so the lookup is testable in one place instead of re-implemented at each call site.

### Three identity/migration findings

- **`ConfigStore.swift:100`** — migration's `resticPath` relocation did a load-mutate-save round trip through the override-aware `MachineStore`, persisting `RESTIC_STATION_MACHINE_ID` into `machine.json`. Now uses a new `MachineStore.persistentIdentity(paths:)` that ignores the variable. Documented in `data-model.md` as the rule for any path that writes `machine.json` for a non-identity reason.
- **`AppModel.swift:148`** — `AppModel` loaded the machine before the config, so the first session after a v1 upgrade missed the `resticPath` the migration had just written. Config is loaded first now; `HelperContext.loadViews` documents and tests the same ordering.
- **`AppModel.swift:232`** — new `machineLoadError`, tracked separately from `configLoadError`, blocks `updateMachine(_:)`. Restic discovery in Settings reaches that path without the user asking to save anything, so the guard belongs in `updateMachine` rather than its callers.

### Regression tests — each verified to fail against the pre-fix code

I reintroduced each bug and confirmed the new tests go red, then reverted:

- `HelperMachineScopeTests` (new, `Helper/Tests/`) — drives `HelperContext.loadViews` against a real `config.json` + `machine.json` in a temp dir. Covers the restore-only host (every set disabled → `scheduled` empty, `addressable` complete), override propagation to utilities, both scopes, `resticPath` in both views, and the load order. Pointing `addressable` at `resolved` fails 7 expectations across 3 tests; swapping the load order fails the migration-visibility test.
- `AddressableResolutionTests` (Core) — the two views, including a **class-of-bug guard**: with every destination overridden, no raw `repoURL` or `nonSecretEnv` may survive into *either* view, so nothing that builds a restic invocation from a `ResolvedConfig` can reach a shared value by accident.
- `ConfigMigrationEnvironmentOverrideTests` (Core, `.serialized`) — sets the **real** environment variable, because injecting it would not reproduce the bug (the code under test reads `ProcessInfo`), and constructs the `ConfigStore` *inside* the override, because `MachineStore` snapshots the environment at construction and a real process is launched with the variable already set. My first draft of this test passed against the buggy code for exactly that reason.

### Verification

- `swift build` + `swift test` (root): green, **16 tests**.
- `swift test --package-path Core`: green, **425 tests in 66 suites**.
- `./scripts/bootstrap.sh && xcodebuild -scheme "Restic Station" build`: green.
- `./scripts/integration-test.sh`: **all assertions passed (46s)** against real restic + keychain, including the v1→v2 migration step.
- CI: `linux` and `macos` both green.

### Not covered by tests

The four app-side call sites are verified by build and inspection only — **this repo has no app test target** (`project.yml` defines the app and the helper tool, no test target for `App/`). The guarantees I could make testable were pushed into Core: the lookup helpers, the two views, and the class-of-bug guard above. The app call sites are now one-liners through those. If an app test target would be welcome, that is worth a follow-up issue.

Existing guarantees re-checked: a config with no overrides resolves to itself in **both** views; migration stays non-destructive and idempotent; resolution still depends only on `machineId` (the cross-platform invariance test now covers the addressable view too, on both platforms).

bherila:
## Fable blocker fixed — `updateMachine` can no longer persist the `machineId` override

Confirmed the finding as reported. Rebased onto `6c3cb6a` (T23).

### The fix

New **`MachineStore.savePreservingIdentity(_:)`** is now the only sanctioned way to persist a non-identity change to `machine.json`:

```swift
let identityStore = MachineStore.persistentIdentity(paths: paths)  // override-blind
let onDisk = try identityStore.load()
var updated = desired
updated.machineId = onDisk.machineId   // re-applied AFTER taking the caller's value
try identityStore.save(updated)
return updated
```

Two properties that matter:

- The identity is restored **after** the caller's value is taken, so it cannot be changed through this method even by accident — and every other field, including ones added to `MachineConfig` later, carries through automatically rather than by an enumerated copy-list that would silently drop new fields.
- It builds on `persistentIdentity` internally rather than on `self`, so it is override-proof **even when called on an override-aware store** — which is exactly the case that produced the bug.

`save(_:)` keeps writing verbatim and is now documented as the identity create/rename primitive.

`AppModel.updateMachine` persists through it and then re-applies the in-memory (possibly overridden) id, so resolution still uses the profile the process was launched with — the two-profile feature keeps working, it just stops leaking to disk.

### Audit of every `machine.json` write path

| Path | Before | After |
|---|---|---|
| `AppModel.updateMachine` | `save(draft)` — **the blocker** | `savePreservingIdentity` |
| `ConfigStore` migration | `save` on a `persistentIdentity` store — already safe | `savePreservingIdentity`, for uniformity |
| `MachineStore.load()` auto-create | `save` with the freshly generated id | unchanged — this *is* identity creation |

Switching the migration was not strictly needed, but it means no reader has to work out which store instance a write is holding to know it is safe. The only remaining `save(_:)` call in production code is the auto-create.

### Tests — red-checked

Four cases in `MachineStoreTests`:

- `savePreservingIdentityKeepsTheOnDiskMachineId` — replays `updateMachine`'s exact sequence: load through an override-aware store (asserting the precondition that it really does see `second-profile`), mutate `resticPath`, save. Asserts the file keeps `studio-mac`, the `resticPath` change *did* land, and the override still applies in memory.
- `savePreservingIdentityIgnoresAnyMachineIdInTheValue` — the guarantee is a property of the method, not of how the caller got its value.
- `savePreservingIdentityCarriesEveryOtherField` — guards against a future field being dropped.
- `savePreservingIdentityCreatesTheGeneratedIdentityWhenAbsent` — no file yet: the identity created is the generated one, never the override.

Red-check: removing the `updated.machineId = onDisk.machineId` line fails all four (6 expectations). Restored, all green.

Also added an integration-test assertion that runs the **real helper binary** under `RESTIC_STATION_MACHINE_ID=second-profile` against a v1 config — so the migration's write to `machine.json` fires while the override is in effect — and confirms the on-disk id is unchanged.

Documented the rule in `data-model.md` §machine.json, naming `savePreservingIdentity` as the required path and `save` as the identity-only primitive, with the reason (the damage outlives the variable, and the write is reached without anyone asking to save).

### Non-blocking fix included

`AppModel.init` now joins the config and machine load diagnoses instead of the machine one overwriting the config one. Accumulated in a local, since `self.configLoadError` cannot be read before all stored properties are initialized.

### Deferred as instructed

Maintenance confirmation over-promising, destination editor showing the raw `repoURL`, and the fixed `machine.json.tmp` temp filename are all left for the milestone-end fix-forward PR.

### Verification

- `swift build` + `swift test` (root): green, **30 tests**.
- `swift test --package-path Core`: green, **472 tests in 69 suites**.
- `xcodebuild -scheme "Restic Station" build`: green.
- `./scripts/integration-test.sh`: **all assertions passed (47s)**, including the new override assertion.
- CI: `linux` and `macos` both green.

Ready for the re-run of the Fable pass.

bherila:
@codex review

PR #38, opened 2026-08-05
T26 — Linux scheduling: systemd user timer

### LINKED ISSUE #28: T26 — Linux scheduling: systemd user timer
# T26 — Linux scheduling: systemd user timer

**Size:** M · **Model:** Sonnet · **Depends on:** T21, T22 · **Milestone:** M5

`docs/scheduling.md` is normative and must be read in full — especially §Tick.

On macOS a launchd agent fires `restic-station-helper tick` **every 2 minutes**, and `App/Sources/Support/LaunchdManager.swift` registers it via `SMAppService`. The tick itself is the entire scheduling contract and is already platform-neutral: it takes `locks/tick.lock`, loads config, recovers interrupted runs, runs due sets sequentially, runs due checks (backup wins), reprobes destinations every 30 minutes, and exits 0 on everything except a hard config-load error. Backups missed while the machine was asleep or off run within one tick of it coming back — anacron-style — because due-ness is computed from `state/schedule-state.json` last-start times rather than from wall-clock cron slots.

**Nothing in `Core/` or the `tick` subcommand needs to change.** This task supplies the Linux mechanism that fires it, and only that. `LaunchdManager` stays macOS-only in `App/` — do not touch it.

## Goal
A Linux user can install a per-user systemd timer in one command, and scheduled backups then run identically to macOS.

### 1. Ship unit templates (`packaging/linux/systemd/`)
- `restic-station.service` — `Type=oneshot`, `ExecStart=<helper-path> tick`. No `Restart=`; a oneshot driven by a timer must not be restarted.
- `restic-station.timer` — `OnBootSec=2min`, `OnUnitActiveSec=2min`, `Persistent=true`, `[Install] WantedBy=timers.target`. `Persistent=true` matters: it makes systemd fire a missed timer after downtime, which is the same anacron-style catch-up macOS gets, and it is what keeps the two platforms behaviourally equivalent.
- **User** units (`~/.config/systemd/user/`), not system units. The helper reads `$HOME`-scoped state and secrets owned by one user; a root-run system unit would break both `AppPaths` resolution and the `0600` secrets file ownership.
- Do not put secrets in an `EnvironmentFile`. Secrets come from T23's secret store. The unit may set `RESTIC_STATION_DATA_DIR` if the user overrode it.
- Templates carry a `<HELPER_PATH>` placeholder substituted at install time — never a hardcoded install prefix.

### 2. `helper timer` subcommand (`Helper/Sources/Commands/Timer.swift`, Linux-only)
- `timer install [--interval <minutes>]` — writes both units to `~/.config/systemd/user/` with the real absolute helper path (resolve via `/proc/self/exe`, not `argv[0]`), runs `systemctl --user daemon-reload` and `systemctl --user enable --now restic-station.timer`. Default interval 2 minutes to match macOS. Idempotent: re-running updates the units in place and does not duplicate anything.
- `timer uninstall` — `disable --now`, remove both units, `daemon-reload`. Idempotent; exit 0 when nothing was installed.
- `timer status` — prints whether the units exist, whether the timer is enabled and active, and the next scheduled firing (`systemctl --user list-timers`), plus the last tick outcome from `state/`. This is the headless equivalent of glancing at the menu bar icon, so make the output genuinely readable.
- Detect and explain the two failure modes that will otherwise generate confused bug reports:
  - **Lingering.** Without `loginctl enable-linger <user>`, user units stop when the user logs out, so scheduled backups silently stop on a headless box. `timer install` must check `loginctl show-user <user> --property=Linger` and, if disabled, print a prominent warning with the exact command to fix it. Do not run it automatically — it needs root, and silently escalating is worse than telling the user.
  - **No systemd.** Detect its absence and fail with a clear message pointing at the cron fallback below, rather than emitting a confusing `systemctl: command not found`.
- Use the existing `ProcessRunning` abstraction for every `systemctl`/`loginctl` call so this is testable with the existing `FakeProcessRunner`.
- Gate the whole subcommand `#if os(Linux)`; it must not appear in `--help` on macOS.

### 3. cron fallback
Document (do not build tooling for) a one-line crontab alternative for non-systemd hosts: `*/2 * * * * <helper-path> tick`. Note the one real difference — cron has no `Persistent=true` equivalent, so a tick missed while the machine was off is not replayed at boot; the next scheduled tick picks it up instead, which delays catch-up by up to the interval but loses nothing, because due-ness comes from `schedule-state.json`, not from cron.

## Tests / verification
- Unit-template rendering is a pure function of (helper path, interval) — test it directly, asserting `<HELPER_PATH>` substitution and that the rendered units parse (`systemd-analyze verify` in CI if available).
- `timer install`/`uninstall` against `FakeProcessRunner`: correct `systemctl` argv sequence, idempotent on repeat, correct behaviour when nothing is installed.
- Linger disabled → warning emitted; systemd absent → clean actionable error.
- Manual end-to-end on a real Linux host (document the transcript in the PR): install the timer, confirm `systemctl --user list-timers` shows it, wait for two firings, confirm `journalctl --user -u restic-station` shows the ticks and that `state/schedule-state.json` advanced.

## Acceptance criteria
- [ ] `swift test` green on macOS and Linux; `timer` is absent from `--help` on macOS.
- [ ] A real Linux host runs scheduled backups end to end after `helper timer install`, with the transcript in the PR.
- [ ] `timer install` is idempotent and warns about linger when it is off.
- [ ] `docs/scheduling.md` gains a Linux section covering the timer, `Persistent=true` as the anacron-equivalent, linger, and the cron fallback with its one behavioural difference.

---

**Blocked by:** #23, #24

bherila:
## What changed

The tick is already the whole scheduling contract and is platform-neutral. This PR supplies the Linux mechanism that fires it, and only that — nothing in `Core/`'s scheduling algorithm or the `tick` subcommand changes.

**`helper timer install|uninstall|status`** (`Helper/Sources/Commands/Timer.swift`, `Helper/Sources/SystemdTimerManager.swift`) — gated `#if os(Linux)`, so it does not appear in `--help` on macOS.

- `install [--interval <minutes>]` writes `~/.config/systemd/user/restic-station.{service,timer}` with the helper's own path resolved from `/proc/self/exe` (never `argv[0]`), then `systemctl --user daemon-reload` followed by `enable --now`. Default interval 2 minutes, matching launchd's `StartInterval`. Idempotent: units are overwritten in place, so re-running with a different `--interval` just changes the interval.
- `uninstall` disables then removes; exits 0 with nothing installed and runs no `systemctl` at all in that case.
- `status` reports units, the interval *actually installed* (read back from the unit rather than assumed), enabled/active, linger state, next firing from `list-timers`, and what the tick has been doing from `schedule-state.json` + `runs/index.jsonl`. Exits 0 only when scheduled backups really will happen, so it works as a health check.

**Unit templates** in `packaging/linux/systemd/` carry `<HELPER_PATH>`, never an install prefix. `Type=oneshot`, no `Restart=`, no `[Install]` on the service; `OnBootSec=2min` + `OnUnitActiveSec=2min` + `Persistent=true` + `WantedBy=timers.target` on the timer. User units only — the helper's state is `$HOME`-scoped and its secrets are `0600`, so a root-run system unit would resolve a different data directory *and* be unable to read the secrets it found. **No secrets in a unit or an `EnvironmentFile`**; the only variable a unit ever carries is `RESTIC_STATION_DATA_DIR`, and only when the installing shell had it set.

**Both confusing failure modes are detected and explained.** Lingering off → a prominent warning with the exact `sudo loginctl enable-linger <user>` line, never run automatically (it needs root). Detection is `loginctl show-user --property=Linger`, falling back to reading `/var/lib/systemd/linger/<user>` — because `show-user` fails outright for a user with no session, which is precisely the headless case the warning exists for. No systemd → detected via the *absence of `/run/systemd/system`* as well as a missing binary (`sd_booted(3)`'s own test: a container can ship `systemctl` while PID 1 is something else), failing with the cron one-liner instead of `systemctl: command not found`, and writing nothing.

**Docs.** `docs/scheduling.md` gains §Linux (units, `Persistent=true`, lingering, the cron fallback and its one behavioural difference, `timer status`), the model section is reframed around the platform split, and §Tick now states explicitly that it is the identical code path both schedulers invoke.

### A finding on `Persistent=true`

Kept as specified, and I did not drop it — but it is worth flagging: `systemd.timer(5)` scopes `Persistent=` to `OnCalendar=` timers, so with the monotonic `OnBootSec=`/`OnUnitActiveSec=` pair specified in the issue it has **no effect of its own**. The post-downtime catch-up this PR actually delivers comes from `OnBootSec=2min`: a machine that was off for a week ticks two minutes after it returns, and that tick finds every set overdue from `schedule-state.json` — behaviourally identical to launchd coalescing missed `StartInterval` fires on wake, which is what the issue asks for. `Persistent=true` stays because it states the intent and is the correct switch the moment the interval becomes a calendar expression; both the rendered unit and the docs now say so inline, so nobody deletes it as dead configuration later. Happy to switch to `OnCalendar=*:0/2` instead if you would rather the directive be live, but that would be cron-slot semantics, which the issue argues against.

## How `LaunchctlCommand` was split

Per the issue comment. `LaunchctlCommand` moves **out** of the portable `Core/.../Support/HelperCommand.swift` into a new `Core/.../Support/SchedulerCommand.swift`, gated `#if os(macOS)`, beside a Linux-gated `SystemdCommand` holding the systemd unit text and `systemctl`/`loginctl` argv.

- **macOS behaviour is byte-identical.** The type, its members and its defaults are unchanged; `LaunchdManager` is untouched. Its tests moved with it (`SchedulerCommandTests.swift`) and still pass verbatim.
- Core no longer compiles launchd vocabulary on Linux, and the split is *stated* rather than implied: the file header names both schedulers, which platform drives each, and where each driver lives.
- `HelperCommand.swift` keeps only what is genuinely portable (the helper's own argv is identical everywhere) and carries a pointer comment to the new file.
- No general scheduler abstraction, no protocol, no shared type — the `#if`s are the whole mechanism. Each platform's driver stays a thin spawn-and-report shell (`App/…/LaunchdManager.swift` on macOS, `Helper/Sources/SystemdTimerManager.swift` on Linux), which is the pre-existing arrangement rather than a new one.

## How this was verified

Locally on macOS (Swift 6.3.3): `swift build`, `swift test`, `swift test --package-path Core` (333 tests), `./scripts/bootstrap.sh && xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO` — all green, and `timer` is confirmed absent from the built helper's `--help`.

Because the development machine is a Mac with no Linux toolchain, the Linux-gated code was additionally compiled and run locally by temporarily flipping every `#if os(Linux)` to `#if true` and running both suites: everything passed except `HelperSelfPathTests`, which needs a real `/proc` and passes in Linux CI. The gates were restored before committing.

`SystemdTimerManager` takes its unit directory, `systemctl`/`loginctl` paths and `ProcessRunning` as injected values, so the real install/uninstall/status code is driven against a scripted `ProcessRunning` double and a temp directory. Covered there: exact `systemctl` argv **and ordering** (`daemon-reload` before `enable --now`; `disable --now` before the files are deleted, because `systemctl disable` reads the unit to find its symlinks); idempotency of both install and uninstall; `0644` unit permissions; `RESTIC_STATION_DATA_DIR` baked in only when overridden; the linger warning in all three states (yes/no/unknown, including the marker-file fallback) and that `enable-linger` is never executed; the systemd-absent error naming the cron line, with nothing written and nothing spawned; a failing `systemctl` surfacing command, exit code, stderr and the sudo/`machinectl` hint for "Failed to connect to bus"; the disable-failed-but-still-remove path; and status output including the INCOMPLETE half-deleted-install case.

Unit-template rendering is a pure function of (helper path, interval) and is tested directly in Core: placeholder substitution, `Type=oneshot`, no `Restart=`, no `[Install]` on the service, interval propagation across `OnBootSec`/`OnUnitActiveSec`/`Description`, `Persistent=true`, quoting of a path containing spaces, and that no unit ever mentions `EnvironmentFile`, `RESTIC_PASSWORD` or "password". A further test asserts `packaging/linux/systemd/*` is byte-identical to what `timer install` writes, so the shipped templates cannot rot.

New Linux CI steps exercise the real binary on a host that genuinely has no systemd (the `swift:6.1` container is not booted with systemd, which makes it an unusually good host for exactly these paths): `timer install` refuses with the cron advice and writes nothing, `timer status` exits non-zero and explains, `timer uninstall` is a clean no-op, and `timer` is asserted present in `--help`. A macOS CI step asserts `timer` is absent from `--help` and that invoking it fails.

## Acceptance criteria — verified vs deferred

- [x] **`swift test` green on macOS and Linux; `timer` absent from `--help` on macOS.** Verified locally (macOS) and in the Linux CI job; both `--help` assertions are now CI steps.
- [ ] **A real Linux host runs scheduled backups end to end after `helper timer install`, with the transcript.** **DEFERRED — cannot be satisfied from this environment.** There is no Linux host, no local Linux toolchain and no container runtime available here; GitHub Actions is the only Linux environment, and its runners are containers with no systemd and no user session bus, so a live `enable --now` + two firings + `journalctl` transcript is not reachable. I have not faked it. Covered by #31 (T29), which adds the Linux integration test, or by a manual pass on a real host.
- [x] **`timer install` is idempotent and warns about linger when it is off.** Idempotency and the warning are verified by unit tests against the scripted runner (repeat install produces the identical argv sequence, exactly two files, updated interval; linger `no` produces the warning with the exact fix command and never runs it). The *live* behaviour of `systemctl enable --now` being a no-op on repeat is systemd's documented contract, not something asserted here.
- [x] **`docs/scheduling.md` gains a Linux section** covering the timer, `Persistent=true` as the anacron-equivalent (with the honest caveat above), lingering, and the cron fallback with its one behavioural difference.

### `systemd-analyze verify`: available, and the units pass

The `swift:6.1` image does not ship it, but `apt-get install --no-install-recommends systemd` inside the container does (systemd 255), so this is **verified in Linux CI**, not skipped: the step renders the shipped templates with `/bin/true` for `<HELPER_PATH>` and both units come back clean. `--user` needs an `XDG_RUNTIME_DIR` to start its throwaway manager, or it dies before reading anything. If a future image cannot install systemd, the step emits a `::warning::` saying the units were **not** parse-verified rather than silently passing.

One thing worth knowing for anyone reading that step: `systemd-analyze verify` **exits 0 even for a directive it could not parse** — it prints `Failed to parse service type, ignoring` and returns success. A negative control (`Type=definitely-not-a-type`) caught this, so the step gates on the diagnostics rather than the exit code, and keeps the negative control so that a clean result cannot be confused with a check that quietly did nothing.

Closes #28


bherila:
@codex review

Issue #29, opened 2026-08-05
T27 — Headless CLI ergonomics: config import/export/validate, status, list


bherila:
# T27 — Headless CLI ergonomics: config import/export/validate, status, list

**Size:** M · **Model:** Sonnet · **Depends on:** T23, T24 · **Milestone:** M5

## Repository context (read first)
`restic-station-helper` (`Helper/Sources/`, swift-argument-parser) today exposes only the subcommands the app and launchd invoke: `tick`, `run-set`, `init-secondary`, `restore`, `probe-repo`, `fda-check`, `version`. They print terse one-line-per-event output because on macOS the **app** is the user interface — the helper is plumbing (`docs/tasks/T10-helper-cli.md`).

Exit-code contract, which every subcommand you add must follow: **0** ok · **1** error · **2** busy (set lock held) · **3** offline (repo unreachable).

On Linux there is no app. The helper *is* the user interface, and the whole point of M5 is that a user authors a config in the macOS GUI and then runs it on a Linux box. Today there is no supported way to get that config onto the Linux host, no way to check it will do what you expect there, and no way to see what happened.

Depends on T24's `machine.json` + per-machine override schema and T23's `secret` subcommands — read both issues before starting; this task is the user-facing surface over them.

## Goal
Close the loop stated in the milestone: author on the Mac, move the config, verify it resolves correctly for *this* machine, and observe the result — all from a terminal.

### 1. `config` subcommand group (`Helper/Sources/Commands/Config.swift`)
- `config export [--out <path>]` — writes the current `config.json` to a path or stdout. Exports the **shared** config only; it must never include `machine.json` or any secret. State this in `--help`, because "export my config" reasonably sounds like it would include the password, and it must be obvious that it does not.
- `config import <path> [--dry-run]` — validates, migrates v1→v2 if needed (reusing T24's migration; do not reimplement it), and atomically installs it. Before overwriting an existing config, back it up beside itself and print a diff summary using the existing `ConfigDiff` (`Core/Sources/ResticStationCore/Config/ConfigDiff.swift`). `--dry-run` prints the summary and exits without writing.
[...]

PR #41, opened 2026-08-05
T27 — Headless CLI ergonomics: config export/import/validate/show, status, sets, runs

bherila:
Closes the loop the M5 milestone states: author a config on the Mac, move `config.json` to a Linux host, verify what it resolves to there, and observe what happened — all from a terminal, with no app involved.

Implements #29 (T27). Depends on #25 (T23) and #26 (T24), both merged.

## Subcommand surface (identical on macOS and Linux)

```
config                  Move the shared config.json between machines, and inspect what it
                         resolves to for one. Never reads or writes machine.json or any
                         secret. Exit 0 ok, 1 error.
  export                Print config.json to stdout, or write it to --out. Never machine.json,
                         never a secret.
  import <path>          Validate, migrate v1→v2 if needed (reusing ConfigStore's own
                         migration), back up any existing config.json, print a change
                         summary. --dry-run writes nothing.
  validate               Hard errors (exit 1); warnings (unrecognized machine ids, cached-
                         unreachable destinations); the effective plan — what runs here and
                         what's excluded, and why. Exit 0 even with warnings.
  show                   The resolved config for one machine, human or --json.

status [--json]          The headless menu bar. Reuses HealthDerivation verbatim. Exit 0
                         healthy/running, 1 on any warning.

sets list [--json]       Every set: id, name, enabled-here, sources, destination count, schedule.

runs list [--json]       Recent runs from runs/index.jsonl.
runs show <runId>        One run's metadata, optionally its full log.
```

## Sample: `config validate` on an all-disabled-here config

This is the anti-silent-failure guarantee the per-machine design (T24) depends on — a set disabled on this machine is never just absent from the report:

```
$ restic-station-helper config validate --machine mirror-box
Errors:
  (none)

Warnings:
  (none)

Effective plan for machine "mirror-box":
  set "Documents" (00000000-0000-4000-8000-0000000000a0) — does not run here
      sources: /Users/bwh/Documents
      schedule: daily 02:30
        - primary "Big Drive": /Volumes/Big/docs.restic  (excluded here)
        - secondary "Scratch HDD": /Volumes/Scratch/docs  (excluded here)

  excluded here, and why:
    - backup set "Documents" is disabled on this machine

  nothing will run on this machine.
```

## Resolution-view choice per subcommand

| Subcommand | View | Why |
|---|---|---|
| `config validate` / `config show` | Both — `.addressable` for inventory (nothing dropped) + `.scheduling`'s omissions for `enabledHere`/exclusion reasons | Reading only `.scheduling` would silently omit an excluded set from the report — exactly the failure T24 exists to prevent. Built once in `EffectiveConfigReport` so the two commands can't disagree. |
| `status` | `.scheduling` | The same view `HealthDerivation` (staleness, health) is documented to run on — matches the app's menu bar. |
| `sets list` | `.addressable`, annotated with `enabledHere` from `.scheduling` | An inventory command must list every set this machine knows about, not just the ones it backs up. |
| `runs list/show` | Neither — reads `runs/index.jsonl`/`runs/<runId>/metadata.json` directly | Runs are historical facts, not resolved-for-a-machine data. |

## `--json` as an interface, not debug output

Every `--json` mode writes only JSON to stdout (piped through `jq` in `scripts/headless-cli-test.sh`); optional fields encode explicit `null`, matching every other persisted-JSON convention in the app (`ConfigStore.makeEncoder()`, now `public`, reused verbatim). Shapes documented in `docs/data-model.md` under "Headless CLI `--json` shapes (T27)". `runs list`/`runs show --json` reuse `RunIndexEntry`/`RunMetadata` verbatim — no new shape to maintain there.

## Secrets

No secret can appear in argv, a log line, a run record, an error message, or any CLI output, in any mode — `config export` explicitly never includes `machine.json` or a stored password (only `Destination.nonSecretEnv` travels). `scripts/headless-cli-test.sh` asserts this by grepping the full combined stdout/stderr of every subcommand it drives against a real fixture secret value with a trailing space and a literal `$` (the same style `secret-cli-test.sh` uses), including through a real `restic init` → `secret set` → `run-set` → `status` round trip.

## Core changes in support

- `ConfigStore.migrateToCurrentVersion` is now `public` and returns `(config, backupWritten)` instead of also installing `config.json` itself — `config import` needs to back up an existing config *before* overwriting it, and `--dry-run` must not install at all. `load()`'s behavior is unchanged byte-for-byte (verified with a regression test plus a manual red-check: temporarily removing the `guard backupWritten` line and confirming the "never overwrite `config.json` without a confirmed v1 backup" invariant test still passed — it did, because both writes share the same directory's write bit under plain POSIX permissions, so that particular test is an end-to-end guarantee rather than an isolated pin; documented honestly in the test's own comment, and the call-boundary itself — that `migrateToCurrentVersion` never installs `config.json` — is pinned separately).
- `ConfigStore.previewMigration` — a pure, side-effect-free version-bump preview for `--dry-run` (no `machine.json` touch, no `config.v1.backup.json` write).
- `ConfigDiff.summarize(from:to:)` — a new human-readable change summary (added/removed/changed sets, `resticPath`), alongside the existing `isScheduleRelevantChange` boolean.
- `StateStore.currentRunSetIDs()` — enumerates live `current-run-*.json` files by filename, the same technique the app's `StateWatcher` already uses, needed for `status`'s `anyRunInFlight`.
- `AppPaths.configImportBackupFile(suffix:)`, and `ConfigStore`'s JSON encoder/decoder/formatter are now `public` so the CLI shares the exact same persisted-JSON conventions instead of reimplementing them.

## Deferred

- `status`'s `backgroundAgentEnabled` input to `HealthDerivation.appHealth` is passed `true` (neutral) rather than checking whether `SMAppService`/the systemd timer is actually registered — documented in code and in `docs/data-model.md`. `timer status` (T26) already covers the Linux half of that question standalone.
- The spec's full macOS→Linux end-to-end demonstration (export on macOS → import on Linux → validate → `secret set` → real backup → `status`) is verified for real on the macOS side (including a real `restic init`/backup round trip through the app-embedded helper) and driven as far as this sandbox's single-host, no-Docker environment allows on the "Linux" side — i.e. the same binary, same code paths, same assertions, but not a literally separate Linux host. The CI `linux` job (swift:6.1 container, no restic installed) exercises every non-restic-dependent assertion for real. Full separate-host verification is tracked on #31 per the task brief.
- `sets list`/`config show`'s human-readable output has no fancy relative-time formatting (`"2h ago"`) — raw ISO 8601 timestamps and second counts only, to keep the formatter surface minimal; a monitoring script or a human can compute that itself.

## Verified locally

- `swift build` && `swift test` (root package) — 56 tests, all passing
- `swift test --package-path Core` — 484 tests, all passing (was 472; +12 from new `ConfigDiff.summarize`/migration-split/`StateStore.currentRunSetIDs` coverage)
- `xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO` — succeeds; bundle layout, helper embedding, and `--help` surface unchanged in shape
- `scripts/integration-test.sh` — passes
- `scripts/secret-cli-test.sh` — passes (regression check; T27 touches no secret-handling code, but shares `ConfigStore`)
- `scripts/headless-cli-test.sh` (new) — passes against both the standalone binary and the app-embedded helper, covering every item in the issue's test list: config import round-trip incl. v1 migration, `config validate` on an all-disabled config (asserted on actual stdout), `config validate --machine <other>`, `status --json` against healthy/in-flight/failed/stale-mirror fixtures with exit-code assertions, every `--json` mode piped through `jq`, and the full-suite secret-leak grep

CI will run both `linux` and `macos` jobs on push — will report status here once green.


bherila:
@codex review

PR #44, opened 2026-08-05
T29 — Static Linux binary, release packaging, Linux integration test

bherila:
## Summary

- Cross-compiles a fully **static** `restic-station-helper` for `x86_64` and `aarch64` using Apple's Static Linux SDK (musl), reproducibly from a clean checkout: `scripts/setup-static-linux-sdk.sh` idempotently installs (no sudo) a pinned swift.org open-source macOS toolchain alongside the matching Static Linux SDK — a same-numbered **Xcode** toolchain is not module-compatible with the SDK's precompiled `Foundation.swiftmodule` ("compiled module was created by an older version of the compiler"), which is exactly what this sidesteps. `scripts/package-linux.sh` builds both architectures, strips debug symbols, verifies static linkage, and packages `restic-station-linux-<arch>.tar.gz` (binary + T26 systemd units + `install.sh` + `LICENSE` + `README.md`) plus `SHA256SUMS`.
- Required `#elseif canImport(Musl)` alongside the existing Darwin/Glibc branches in 8 files (the musl SDK exposes libc as `Musl`, not `Glibc`) — the only source changes the static build needed. No Foundation-on-musl rough edges surfaced in the code paths exercised (Process, FileManager, flock, file I/O).
- **Extends** `scripts/integration-test.sh` (not a Linux sibling) to run on both platforms: an OS-conditional secret backend (keychain on macOS, `FileSecretStore` on Linux), plus a new M5-story scenario (`assert_fixture_flow`) that imports a **checked-in fixture** shaped like a real macOS `config export` (`scripts/fixtures/mac-exported-config.json`), sets a secret via stdin, runs a real backup + a second destination as a mirror (`restic copy`), and asserts on actual `config validate`/`status --json` output that a set disabled for this machine is correctly excluded (the anti-silent-failure guarantee of the per-machine design).
- CI: `release-linux` builds + packages + uploads both architectures; `linux-runtime-verify` proves genuine staticness (`ldd`, `scratch`, Alpine, Debian 12, and — x86_64 only — CentOS 7 as the deliberately old-glibc distro); `linux-integration` runs the extended integration test against the **static** binary on a real (non-container) Ubuntu aarch64 VM with real restic installed, which is what finally provably closes T23's "`RESTIC_PASSWORD_COMMAND` against `FileSecretStore` on Linux" criterion.
- `docs/testing.md` and `README.md` updated; `packaging/linux/README.md` is the tarball's own first-run doc.

## Verification debt (issue #31's own tracking)

- **T23** — closed by `linux-integration`: real restic is installed there and the extended integration test runs the full `RESTIC_PASSWORD_COMMAND`/`FileSecretStore` path on Linux end to end.
- **T26** — still deferred. `linux-integration`'s bare Ubuntu VM (systemd really is PID 1 there, unlike the `swift:6.1` container) gets a non-blocking diagnostic `timer install`/`timer status` step so the PR's CI logs show the actual current state rather than an assumption, but a GitHub-hosted runner's default account has no logind session, so this is expected to still report "no systemd user bus." Needs a real host (or self-hosted runner with a login session) to close out.

## Test plan
- [x] Local sweep green: `swift build`, `swift test`, `swift test --package-path Core`, `xcodebuild`, `./scripts/integration-test.sh` (macOS, keychain + fixture flow), `./scripts/headless-cli-test.sh`.
- [x] `scripts/package-linux.sh` run locally (macOS → both Linux archs): both binaries `ELF … statically linked`, tarballs + `SHA256SUMS` produced, `install.sh` verified idempotent/unprivileged/PATH-warning by hand.
- [x] Red-check: the `config validate` disabled-set-exclusion assertion was deliberately broken (flipped the fixture's `enabled: false` → `true`) and confirmed to fail at that exact step, then restored and re-verified green. Two more bugs were caught the same way during development (a missing `RESTIC_STATION_DATA_DIR`/`RESTIC_STATION_SECRET_BACKEND` forward in a helper function, and an uppercase/lowercase UUID mismatch against `runs/index.jsonl`) — both reproduced with the exact failure before the fix and confirmed green after.
- [ ] CI green: `linux`, `macos`, `release-linux`, `linux-integration`, `linux-runtime-verify` — driving now.


bherila:
@codex review

bherila:
Checked this against the pinned toolchain rather than the help text, and the premise doesn't hold on Swift 6.3.3 — `swift sdk list` prints the *bundle* name there:

```
$ swift sdk list
swift-6.3.3-RELEASE_static-linux-0.1.0
```

which is exactly what the check greps for, so a second run takes the already-installed branch and the script is idempotent as advertised.

Hardened it anyway in the following commit, because the underlying reasoning is right: the output shape isn't a contract, `--help` really does say "IDs", and `SWIFT_VERSION` is deliberately overridable — so pinning the check to one toolchain's formatting is a latent break for exactly the people this script exists to serve. It now accepts either the bundle name or both per-target ids (requiring *both* arches, so a half-installed bundle still reinstalls), and a `swift sdk install` failure no longer aborts when the SDK turns out to be present anyway.

Issue #46, opened 2026-08-06
timer status exits 0 even when lingering is disabled


bherila:
## What's wrong

\`SystemdTimerManager.status(...)\` (\`Helper/Sources/SystemdTimerManager.swift\`) computes its
boolean return value — which becomes \`timer status\`'s exit code — as:

\`\`\`swift
return hasService && hasTimer && enabled == "enabled" && active == "active"
\`\`\`

Lingering state is fetched (\`lingerState()\`) and printed in the human-readable report ("linger
DISABLED — scheduled backups stop when \<user\> logs out"), but it plays no part in the exit
code. A host with lingering disabled — the single most common way a Linux timer install quietly
stops working after the admin's SSH session ends — reports \`timer status\` exit 0 as long as the
unit is enabled+active, even though the next logout will kill it.

\`timer status\` is documented (\`docs/scheduling.md\` §\`timer status\`, \`docs/linux.md\`
§Troubleshooting) as "exits 0 when scheduled backups will happen and 1 otherwise, so it works as
a health check" — a disabled-linger host does not meet that bar today.

## Suggested fix (not scoped here)

Fold \`LingerState.disabled\` into the health boolean, or add a distinct exit code for "installed
and active but will not survive logout." Needs a decision on whether \`.unknown\` (loginctl
unreachable, e.g. inside the \`swift:6.1\` CI container) should also downgrade the exit code, or
only a confirmed \`.disabled\`.

## Why not fixed in T30

Found and documented (not fixed) while writing \`docs/linux.md\` (T30, #32) — a docs task is the
wrong place to change scheduling/exit-code behavior. \`docs/linux.md\`'s troubleshooting section
tells readers to check the \`linger\` line explicitly rather than trust the exit code, and links
here.

bherila:
Related finding from `@codex review` on #47, same failure class from the other direction:

**`status --json` cannot see the scheduler at all.** `Status.run()` passes `backgroundAgentEnabled: true` into `HealthDerivation` unconditionally, and on Linux the scheduler's real state is exposed only by `timer status`.

So the two documented health checks have complementary blind spots:

| check | blind to |
|---|---|
| `status --json` | timer absent, disabled, or inactive — exits 0 regardless |
| `timer status` | lingering disabled (this issue) — exits 0 regardless |

A monitoring setup running either one alone stays green while scheduled backups are stopped, until a set eventually goes stale. Running both still misses the linger case until this issue is fixed.

[...]

Issue #45, opened 2026-08-05
Verify a live systemd --user timer end to end on a real Linux host (T26 deferred criterion)


bherila:
## What is unverified

T26 (#28) installs a systemd `--user` timer via `helper timer install`. The unit generation,
the no-systemd fallback, and `timer status` are all covered by CI, but **a live timer actually
firing `tick` on a schedule has never been observed.**

Deferred from T26, carried through T29 (#31), and now tracked here so it does not disappear
into a closed issue.

## Why CI cannot cover it

A GitHub-hosted runner's default account has no logind session, so there is no user D-Bus bus
for `systemctl --user` to talk to. T29's `linux-integration` job runs on a bare Ubuntu VM where
systemd genuinely is PID 1 — unlike the `swift:6.1` container used by the `linux` job — and it
carries a `continue-on-error` diagnostic step running `timer install` / `timer status` so the
CI logs show the real state rather than an assumption. As expected, it reports no systemd user
bus.

## What *is* covered today

- Unit file generation, including the `<HELPER_PATH>` substitution.
- `systemd-analyze verify` gating on **diagnostics**, not exit code — it exits 0 on unparseable
  directives — with a negative control proving the gate works.
- The no-systemd path: `timer install` refuses cleanly and prints a cron fallback line.
- `timer status` output shape.

## How to verify

On a real Linux host with a login session (or a self-hosted runner with lingering enabled):

```sh
loginctl enable-linger "$USER"      # survive logout — required for a headless box
restic-station-helper timer install
systemctl --user list-timers | grep restic-station
# wait for at least two firings, then:
restic-station-helper runs list --limit 5
journalctl --user -u restic-station.service --since '-30min'
```

Confirm: the timer fires on the expected cadence; `OnBootSec=2min` produces the catch-up run
after a reboot; and due-ness still comes from `state/schedule-state.json` rather than the timer
cadence, so a missed window is caught up rather than skipped.

## Notes

- `Persistent=true` in the unit is **inert** — `systemd.timer(5)` scopes it to `OnCalendar=`
  timers, and this unit uses the monotonic `OnBootSec=`/`OnUnitActiveSec=` pair. It is kept and
  documented inline so nobody deletes it as dead config; real catch-up comes from `OnBootSec=2min`
  plus the schedule state. Worth re-confirming during this verification.
- Related deferred verification: #43 (FDA attribution through the macOS CLI symlink).

bherila:
[...]

PR #47, opened 2026-08-06
T30 — Documentation: docs/linux.md and cross-platform doc sweep

bherila:
## Summary

- Adds `docs/linux.md`: the end-to-end Linux headless guide (download → install → move a config
  from macOS → per-machine setup → secrets → scheduling → operating it → troubleshooting).
  Every transcript in it is real CI output, not paraphrased or relabeled from a macOS terminal —
  see "Real vs. not-executed" below for exactly which command came from which job/step.
- Cross-platform doc sweep: `README.md`, `docs/architecture.md`, `docs/ui-spec.md`,
  `CONTRIBUTING.md`, `docs/testing.md`, `docs/release.md`.
- New `scripts/linux-docs-transcript.sh` + a CI step in `linux-integration`
  (`.github/workflows/ci.yml`) that runs the exact commands `docs/linux.md` quotes against the
  **static** release binary and real restic 0.18.1, on a genuine (non-container) Ubuntu VM. This
  makes the docs regression-checked going forward: if a command's output changes, this step's log
  changes with it.
- Along the way, found and closed a real verification gap: [issue #45](https://github.com/bherila/restic-station/issues/45)
  assumed a GitHub-hosted runner has no logind session at all. That turned out to be wrong for the
  `linux-integration` runner — `loginctl` reports a real session with lingering enabled — so the
  script now waits (up to ~3 min) after `timer install` and confirms a **real, unattended systemd
  timer firing** that runs a genuine backup (`"trigger":"scheduled"`, `journalctl` shows the
  service starting/finishing). Issue #45 is updated with the finding and left open for what a
  single CI job still can't show (an `OnUnitActiveSec=` repeat, surviving an actual logout).
- Filed [issue #46](https://github.com/bherila/restic-station/issues/46): `timer status` exits 0
  even when lingering is disabled (confirmed by reading `SystemdTimerManager.status`'s return
  expression — linger state is fetched and printed but never folded into the exit code).
  `docs/linux.md`'s troubleshooting section tells readers to check the `linger` line explicitly
  rather than trust the exit code; not fixed here, since a docs task is the wrong place to change
  scheduling exit-code behavior.

## The audit (per the task's real deliverable)

Doc by doc, what T22–T29 got right, what was missing/wrong, and what I did:

| Doc | Audit finding | Action |
|---|---|---|
| `docs/testing.md` §Layer 2 | The task's own "known drift" list said this still describes the integration script as macOS-only. **Already false as of T29** (b8265c5, already on `main`) — it already says "Runs on **both platforms**" and explains the Linux extension in detail. | No fix needed; reported here so the audit doesn't silently skip a claim that turned out already resolved. Added a new subsection documenting the new `linux-docs-transcript.sh` step and updated the CI job table for it. |
| `docs/ui-spec.md` | Confirmed gap: never stated it's macOS-app-only, despite being entirely SwiftUI/AppKit-specific screen specs. | Added a lead sentence: "This document is macOS-app-only... nothing on Linux is expected to match anything described here." |
| `README.md` | Confirmed gap: opened with "A native macOS menu bar app," Requirements section was macOS-only, components table had no platform column — even though a whole "Linux (headless)" section already existed further down (added by T29), creating internal contradiction. | Rewrote the opening line, split Requirements into macOS/Linux, added a Platform column to the components table, added 5 FAQ entries the design provokes (Linux-without-Mac, one-config-multi-machine, why-no-Linux-GUI, repo-interchangeability, Linux secret storage), added `docs/linux.md` to the doc table. |
| `docs/architecture.md` | Confirmed gap, though not quite as described: the doc has no markdown *table* for Components at all (a numbered prose list instead), so "the components table has no platform column" meant there was no table to annotate. | Added a Component/Platform/Role table above the existing prose list; noted the process diagram is macOS-specific and pointed to the Linux equivalent. |
| `docs/scheduling.md` | Audited T26's Linux section closely (linger handling, the two systemd units, `Persistent=true` inertness, cron fallback). All accurate and already cross-references the exit-code caveat in prose — but I found, by reading `SystemdTimerManager.swift` directly, that `timer status`'s **exit code** doesn't reflect the documented "0 when scheduled backups will happen" claim when lingering is off. Not a doc gap — a code/doc mismatch. | Filed issue #46 (see above); doc already correctly warns to check the `linger` line, but I strengthened `docs/linux.md`'s troubleshooting to state the exit-code caveat explicitly and not just imply it. Left `scheduling.md` itself unchanged — it doesn't claim an exit-code guarantee as strongly as `timer status --help`'s own abstract does. |
| `docs/keychain-and-fda.md` | Audited T23's §5 (Linux `secrets.json` backend) in full: threat model, `0600`/`0700` rules, atomic writes, concurrency, argv-safety. All correct and thorough — verified the worked troubleshooting messages against `FileSecretStore.swift` and CI output; matched. | No changes needed. |
| `docs/data-model.md` | Audited T24's per-machine scoping section (schema v2, `machine.json`, the two worked examples, resolution algorithm) against `Core/Sources/ResticStationCore/Config/Models.swift` and both fixtures (`config-v2.json`, `mac-exported-config.json`). Everything matches the shipped code exactly — the two worked examples are drawn directly from the real fixture and I re-validated them for real in CI rather than trusting them by inspection. | No changes needed; reused these exact worked examples (with attribution) in `docs/linux.md`. |
| `CONTRIBUTING.md` | Confirmed gap: said Core "must compile and pass its tests on Linux" but never mentioned `Helper/` (which also must build on Linux — it's in the same root SwiftPM package the `linux` CI job builds), and referred to "CI must be green" without naming which jobs. | Rewrote the build/test paragraph to name `Helper/` explicitly and all five CI jobs by name; added the Linux secrets/`machine.json` invariants to the "key invariants to preserve" list. |
| `docs/release.md` | Not in the task's required list, but found during the sweep: the pre-flight checklist named CI jobs `core-linux` and `macos` — `core-linux` hasn't existed since before the Linux port; the real jobs are `linux`/`macos`/`release-linux`/`linux-integration`/`linux-runtime-verify`. | Fixed the job names. Did not attempt a fuller "how to also cut a Linux release" section — out of scope for this task and would need its own signoff on what "a Linux release" even means for an unsigned static binary; noted here rather than silently expanded. |
| `docs/restic-cli.md` | The `mount` section is macOS-specific (macFUSE, `diskutil`) with no Linux equivalent mentioned. | **Deliberately left alone** — not in the task's required doc list, and `docs/architecture.md`'s `AppPaths` table already notes `mount` is "macOS-only in practice" (headless Linux generally lacks FUSE). Documenting a feature that isn't really usable on the target Linux hosts felt like scope creep. |

## Real vs. not-executed

Every transcript in `docs/linux.md` is one of:

1. **Generated by `scripts/linux-docs-transcript.sh`**, run in `linux-integration`'s "Docs
   transcript for docs/linux.md" step — config import/validate, secrets (set/list/permission
   widening), repo init, `run-set`, `status`/`runs list`/`runs show --log`, `restore`,
   `probe-repo` (reachable + offline), the two per-machine worked examples re-validated against
   the checked-in fixture, `timer install`/`status`, and the real timer-firing wait/verify. See
   e.g. https://github.com/bherila/restic-station/actions/runs/31060708865/job/92488175606.
2. **Pulled from other already-existing CI steps**, cited by job/step name in the doc: `install.sh`'s
   transcript from `linux-integration`'s own "`install.sh` works unprivileged..." step; the cron
   fallback message and the `ResticDiscovery` "restic not found" message from the `linux` job's
   `swift:6.1`-container steps (the natural place those specific messages show up for real, since
   that container genuinely has no systemd and no restic).
3. **Marked "not executed here"**, in `docs/linux.md`'s own "Not executed in CI" section, with a
   reason and a cross-reference: the interactive echo-disabled `secret set` TTY prompt (no TTY in
   CI), and an `OnUnitActiveSec=` repeat firing / surviving an actual logout (→ issue #45, updated
   with what *was* newly observed).

Nothing was typed by hand on this Mac and presented as Linux output.

## Verification

- `swift build`, `swift test` (root package, 66 tests) — pass.
- `swift test --package-path Core` (518 tests) — pass.
- `xcodegen generate` + `xcodebuild -scheme "Restic Station" build CODE_SIGNING_ALLOWED=NO` — pass.
- `./scripts/integration-test.sh`, `./scripts/headless-cli-test.sh` (real restic on PATH,
  `RESTIC_STATION_DATA_DIR` isolated) — pass.
- CI: all six jobs green (`linux`, `macos`, `release-linux`, `linux-integration`,
  `linux-runtime-verify` ×2 architectures) on the final commit — see checks below.

## Deliberately left alone

- `docs/restic-cli.md`'s `mount` section (see audit table above).
- A fuller Linux section in `docs/release.md` beyond the stale job-name fix.
- Issue #46 (`timer status` exit-code/linger bug) — documented as a known issue with a workaround, not fixed; a docs task shouldn't change scheduling behavior.
- Issue #45 is **updated**, not closed — a live unattended firing is now observed for the first time, but repeat firings and logout-survival remain genuinely unverified by any single CI job.

**Do not merge** — awaiting `@codex review` and safety review per the task instructions.


bherila:
@codex review

PR #51, opened 2026-08-06
Health checks that tell the truth

### LINKED ISSUE #46: timer status exits 0 even when lingering is disabled

### LINKED ISSUE #48: timer install does not preserve a custom XDG_STATE_HOME — unattended ticks can silently target the wrong data dir
## The failure

`helper timer install` bakes `RESTIC_STATION_DATA_DIR` into the generated systemd unit when it is
set, but **not** a custom `XDG_STATE_HOME`.

A systemd `--user` service receives the *user manager's* environment, not the interactive shell's.
So for a user whose shell profile sets `XDG_STATE_HOME=/srv/state` (and who therefore has their
config, secrets and run history under `/srv/state/restic-station`):

1. `timer install` from that shell writes a unit with no `XDG_STATE_HOME`.
2. After logout or reboot, the tick runs with the manager's environment.
3. `AppPaths.default()` falls back to `~/.local/state/restic-station`.
4. It finds no `config.json`, no secrets — and **backs up nothing, silently**.

Backups stop and nothing reports it. That is this project's worst failure class, and it is
indistinguishable from "nothing was due" to anyone watching.

Found by `@codex review` on #47.

## Why the current docs are only a workaround

`docs/linux.md` (as of #47) tells readers with a custom `XDG_STATE_HOME` to pass
`RESTIC_STATION_DATA_DIR` explicitly at `timer install` time, which *is* persisted. That is correct
guidance and unblocks the docs, but it relies on the user reading a caveat to avoid silent data loss.

## Suggested fix

`timer install` should resolve the data directory it is *actually* running against — via `AppPaths`,
which already resolves the override chain — and persist that resolved absolute path into the unit
unconditionally, rather than only forwarding an env var it happens to see. The unit should never
depend on the tick re-deriving a path from an environment it will not have.

Worth checking the same question for the cron fallback path, which had a related bug
(`RESTIC_STATION_DATA_DIR` dropped from the suggested crontab line) already noted in the M5
fix-forward backlog.

## Related

- #46 — `timer status` exits 0 even with lingering disabled (another way scheduling can be broken
  while the health check stays green).
- #45 — live systemd timer verification (a real unattended firing *is* now confirmed in CI; long-horizon
  and logout survival are not).

bherila:
Six defects, one class: **this host reports healthy while scheduled backups have stopped.** On a headless box that is the difference between a backup system and the appearance of one — and `docs/linux.md` had to ship every one of these as a caveat telling readers not to trust the exit code in front of them.

Closes #46, closes #48. Also fixes fix-forward backlog items 10, 11, 21 and 22.

## What was broken

| where | it reported | reality |
|---|---|---|
| `timer status` | exit 0 | lingering disabled — systemd kills these units at the next logout |
| `timer status` | exit 0, "no backup sets configured" | `config.json` will not parse; every tick exits 1 |
| `timer install` | success | a custom `XDG_STATE_HOME` never reached the unit; the tick backs up nothing, silently |
| `status` | exit 0, idle | `runs/index.jsonl` unreadable — healthy *because* it could not read the evidence |
| `status` | exit 0, running, forever | one `SIGKILL`ed run left a `current-run` file nothing ever deletes |
| `status --json` | no scheduler field at all | `backgroundAgentEnabled: true` hardcoded on every platform |

## Notable decisions

**`.running` still outranks `.warning`.** That is settled semantics shared with the macOS menu bar, and it is not inverted here. What changed is that "in flight" now has to be *true*: liveness comes from the run's recorded `pid` via `runs/<runId>/metadata.json` — the same evidence `recoverInterrupted()` already used, so the two can never disagree about whether a run died.

**Deliberately not a staleness heuristic on `updatedAt`.** Progress is written only when restic emits a `status` line, and some phases legitimately emit nothing for hours (`check --read-data` writes one phase marker and goes quiet). Any "no progress for N minutes ⟹ dead" threshold either cries wolf on those runs or is set so high it stops being a health check. Process liveness has neither failure mode. There is a test asserting exactly this: a *fresh* `updatedAt` from a dead process is still wreckage.

**Two unknowns are kept as unknowns, not as findings.** A permanently-red health check is one nobody reads, so:
- linger `unknown` (no logind — a container) does not fail `timer status`; only a confirmed `Linger=no` does.
- `status` on a host with no systemd reports `scheduler: {"kind": "unknown", "healthy": null}`. The documented fallback there is a cron line, which nothing can inspect. `backgroundAgentEnabled` is `Bool?` so this and macOS's app-only `SMAppService` state both contribute nothing — the same rule an absent `fda-check.json` already followed.

**The cron fallback got the same pin, and needed it more.** cron sources no profile at all. Also `sh`-quotes the paths and escapes `%` — `crontab(5)` turns a bare `%` into a newline, silently scheduling a *shorter* command than the one printed.

## Verification

**Every regression test red-checked** — bug reintroduced, test confirmed failing, source restored. 8 unit-level plus both new integration assertions. Nothing was left mutated; the tree diff was checked afterwards.

**The Linux-only code was actually compiled and run.** It is `#if os(Linux)` and so invisible to a build on this machine. I built and ran the whole tree under a forced-Linux copy (`#if os(Linux)` → `#if true`): 103 tests, one failure, and it is `resolvesTheRunningBinary` needing a real `/proc/self/exe`. Everything Linux-specific in this PR passes there. That is a compile-and-logic check, **not** a claim about real systemd — that is CI's job, below.

**What CI proves, not just prints.** `scripts/linux-docs-transcript.sh` now fails the job with an explicit `REGRESSION:` message if the unit loses its pinned data directory (#48), if `status --json` stops reporting `scheduler.healthy == false` on a host whose timer is gone (#46), or if an abandoned run stops being reported. Two new `linux`-job steps cover the container case — `status` must **not** go red merely for lacking systemd, and must exit non-zero when the run index is unreadable (using a directory-as-file so the check is not a no-op under root).

**`docs/linux.md`'s five affected transcripts are PLACEHOLDER blocks**, to be filled from this PR's own CI run. Not hand-written: the document's opening claim is that every transcript in it is real output, and I am not going to weaken that by predicting what the new output looks like.

## Also verified

- Full `scripts/integration-test.sh` passes on macOS (63s), including the two new assertions.
- macOS app builds (`xcodebuild -scheme "Restic Station"`).
- Core 528 tests / Helper 66 tests green on macOS; Core 541 / Helper 103 under forced-Linux.


bherila:
@codex review

Focus areas, since this PR is entirely about not lying to a monitoring script:

1. **Is any of the new health logic itself capable of reporting healthy while backups are stopped?** That is the defect class being fixed; a partial fix that closes five of six holes is the worst outcome here.
2. **`RunStore.liveness(ofCurrentRun:)`** — the pid-liveness decision. In particular: the missing/undecodable-metadata case is treated as `.abandoned`, on the reasoning that `begin()` writes `metadata.json` before the first progress write. Is there any ordering or partial-write window where a genuinely live run would be misread as wreckage?
3. **`Tick.clearAbandonedProgress`** — it deletes a state file. The `runId` guard is the only thing stopping it from deleting a *newer* run's live progress for the same set. Is that guard sufficient, and is the read-then-delete race benign?
4. **Three-valued `backgroundAgentEnabled: Bool?`** and the two unknown paths (macOS `null`, no-systemd `kind: "unknown"`). Does `nil` ever silently become `true` again through `scheduler.flatMap(\.healthy)`?
5. **`SystemdCommand.cronFallbackLine`** — new shell quoting and `%` escaping. Is the `'\''` handling right, and is escaping `%` *after* quoting the correct order given cron strips the backslash before `sh` sees the line?
6. Whether `status` spawning `systemctl` is safe for a command documented as a cheap read-only monitoring check.

Please verify each finding against the source before reporting it.

bherila:
Thanks — 8 of 11 fixed in bffb6ee. Two of them were the same defect this PR is about, arriving through doors I hadn't checked, and one would have shipped a wolf-crying regression. Detail below, including the three I'm deferring and why.

## Fixed

**Fail status when a live run masks a broken scheduler** (P1) — confirmed and fixed. `appHealth` is unchanged (`.running` outranking `.warning` is settled semantics shared with the macOS menu bar), but `status` no longer exits on `health == .warning`. It exits on a new `HealthDerivation.hasWarningConditions` — the same rules minus the precedence. Both read one shared predicate, and there's a test that walks every warning input through both and asserts they agree, so "what counts as a problem" can't fork.

**Do not mark finishing live runs as abandoned** (P2) — confirmed, and in effect the most serious of the eleven. Verified the ordering you describe: `performChild` → `RunStore.finish` moves the child's metadata off `.running`, while `current-run` is cleared only by the set-level `defer` at `BackupEngine.swift:203`, with the next child's phase marker rewriting it in between. `liveness` is now pid-only. Crying wolf mid-backup is exactly how a health check teaches people to ignore it, so thank you for this one.

**Escape systemd specifiers in the pinned data directory** (P1) — fixed, `%` → `%%`, and `installedDataDirectory` undoes it on read so a `%` path doesn't report a permanent false mismatch instead.

**Verify the timer targets the status data directory** (P1) — fixed. The unit's pinned value is compared against the resolved one; a mismatch and a unit that pins *nothing* (i.e. written before #48) are both findings now, with distinct `TimerProblem` cases.

**Bound the entire scheduler probe** (P2) — fixed. `status` passes `verdictOnly: true`, which drops `list-timers` (pure narrative for a caller that discards the log) and bounds the remaining calls at 5s each.

**Recover abandoned runs before configuration early exits** (P2) — fixed. Hoisted to step 2; it only ever needed `RunStore`/`StateStore`. You're right that the early-exit states are precisely where wreckage strands.

**Make the guarded current-run deletion atomic** (P2) — fixed under the set lock, which inverts the question usefully: a live run holds that lock for its whole duration, so failing to acquire it *is* the answer.

**Shell-quote the abandoned-run removal command** (P2) — fixed; quoting moved to a shared `ShellQuoting` so the cron line and this use one rule.

## Deferred

**Account for backslashes when escaping cron percent signs** (P2) — I believe you, but I won't guess. The fix depends on Vixie cron's scanner treating `\\%` in a way I can't verify from here: no Linux host, and cron isn't in the CI container either. Guessing at an escaping parity rule and shipping it is worse than the current state, which is only wrong for a path containing a literal backslash immediately before a percent. Filing it with your analysis attached.

**Do not equate an existing helper PID with a healthy run** (P1) — correct, and out of scope. `SIGSTOP`/deadlock defeats `kill(pid, 0)`, and the honest fix is a heartbeat or watchdog — a design change, not an edit. Worth noting the same limitation already exists in `recoverInterrupted()`, which this PR did not introduce; making them both robust is one coherent piece of work. Filing it.

**Do not turn an unknown scheduler into exit success** (P1) — this one is a deliberate decision, not an oversight, and I'd like to keep it while acknowledging your point is real. `unknown` covers a macOS host (SMAppService is app-only) and a Linux host with no systemd, where the documented fallback is a cron line nothing can inspect. Treating those as non-success makes `status` exit non-zero on *every* macOS host and *every* container, forever, with no reachable fix — and a permanently-red check is one nobody reads, which is the same failure in a different costume. The genuinely correct fix is to stop it being unknown: read `launchctl print gui/$UID/<label>` on macOS. Filing that instead, since it closes the hole rather than reclassifying it.

## Note on verification

GitHub Actions has been in a major outage for the duration of this review, so none of this has CI behind it yet. What it does have: 8 more red-checks (bug reintroduced, test confirmed failing, source restored), and the red-check harness now refuses to score one unless the filter first matches a *passing* test — which immediately caught a false negative of mine where a Linux-gated test simply hadn't run on macOS.

Also added `scripts/local-ci.sh`, which runs everything runnable on a Mac including a real `os(Linux)` cross-compile via the Static Linux SDK, and prints exactly what it cannot cover. It earned its keep on the first run by catching a regression neither of us flagged: `headless-cli-test.sh`'s in-flight fixture fabricated a `current-run` file with no run record behind it — which, after this PR, is indistinguishable from wreckage. That fixture only ever passed because nothing checked. It now writes a live run record, and has an abandoned counterpart asserting the opposite.

bherila:
## Fable safety review

Verdict: **merge after addressing Finding 1** — fixed in the commit above. Findings 2 and 3 filed as #57 and #59.

**Finding 1 (PROVED, fixed here).** `recoverInterrupted()` only returns runs it itself transitioned from `running` to `failed`, and tick cleared current-run files only for what it returned — so wreckage whose metadata was *already* terminal was returned by nothing and cleared by nothing. I verified both halves against the source before fixing: `RunStore.swift:247`'s `guard metadata.status == .running`, and `hasWarningConditions` counting abandoned runs directly from `runsInFlight` while `sets[]` is built only from the resolved config.

The reason this matters more than the odds suggest: the state arises routinely on **every host upgrading from the previous release**, where the first post-kill tick already rewrote metadata to `failed` while nothing deleted the current-run file. The hosts this PR exists to rescue are the ones most likely to be in it. If the set had since been removed from the config, `status` exited 1 forever while naming nothing — a permanent alarm with no stated reason, which is the crying-wolf route back to the false assurance this PR is about.

Tick now sweeps every current-run file whose liveness is `.abandoned`, still under the set lock. The `runId` match is replaced by the liveness check, which is strictly stronger under the lock. Red-checked in `scripts/integration-test.sh`: with the old condition restored, the new assertion fails.

The other half of Finding 1 — surfacing an *unattributed* abandoned run in `status` output — is **#56** rather than widened into this PR. The sweep bounds that window to one tick instead of forever, but a `status` inside the window still gets an undiagnosable alarm.

**Also filed:** **#58**, a residual Fable identified that this PR does not introduce — a configured set that has *never* run trips nothing on a host whose scheduler cannot be inspected, because staleness (the backstop that time-bounds #54/#55) exempts never-synced destinations by design. Fresh install + dead agent = exit 0 indefinitely.

**Verified clean and worth recording**, since it scopes the bill of health: the `begin()`-writes-metadata-before-progress ordering claim holds (atomic temp + rename); pid-only liveness is correct for multi-child runs in both directions; every `BackupEngine` entry point holds the same set lock for its whole duration, so nothing deletes state a live run owns; each exit-code hole (unreadable config, unreadable machine identity, unreadable index, abandoned run, disabled/missing/inactive timer, linger off, data-dir mismatch, live-run-masks-broken-scheduler) exits 1; no secrets in unit files, cron lines, or new output; and the load-bearing tests are not vacuous — the CI assertions check JSON the old code never emitted, in both the false-assurance and crying-wolf directions.

PR #60, opened 2026-08-07
Enforce 0700 mode for existing secret directories

### LINKED ISSUE #49: FileSecretStore's 0700 directory mode is not enforced when the data dir already exists
## What is actually guaranteed

- **`secrets.json` is created `0600`** — via `open(O_CREAT|O_EXCL|O_WRONLY, 0o600)` with a preceding
  `unlink` so `O_EXCL` always creates a brand-new file (`open(2)` ignores `mode` for an existing
  file). This is real and enforced. It is also the protection that matters, since it is what keeps
  the secret unreadable.
- **The containing directory is `0700` only when `FileSecretStore` is what created it.**

## The gap

`FileSecretStore.createDirectory(at:mode:)` (`Core/Sources/ResticStationCore/Secrets/FileSecretStore.swift:462-483`)
returns early if the directory already exists:

```swift
var info = stat()
if url.path.withCString({ stat($0, &info) }) == 0 {
    return
}
```

In the normal flow the directory *does* already exist: `AppPaths.ensureDirectories()` creates `root`
through `FileManager.createDirectory`, which uses the process default mode — typically `0755`. By the
time a secret is stored, `prepareDirectories()` finds `root` present and leaves it at `0755`.

**Confirmed empirically**, not inferred: the `linux-integration` transcript captured for `docs/linux.md`
in #47 reports `755 .` for the data directory on a real Linux host.

## Impact

Modest but real on a multi-user host: other local users can `ls` the directory and see which
destination IDs have stored secrets, and the sizes/mtimes of the store. They cannot read the secret
itself — `secrets.json` is `0600`.

The larger problem is that the guarantee was **documented and repeated as stronger than it is**
(`docs/linux.md`, `CONTRIBUTING.md`), which is corrected in #47.

## Suggested fix

Have `prepareDirectories()` tighten an existing directory to `0700` when it will hold secrets, rather
than returning early — while keeping the create-at-mode path for the fresh case, so there is still no
window where the directory exists at a looser mode. Consider whether `AppPaths.ensureDirectories()`
should create `root` at `0700` in the first place, given everything under it is this app's private
state.

Add a regression test asserting the mode after the `ensureDirectories()`-then-`secret set` ordering
specifically — the ordering is what defeats the current code, so a test that only exercises the fresh
path will pass while the bug is present.

Found by `@codex review` on #47 and verified against the source.

bherila:
## Summary

- create a fresh app data root with owner-only `0700` permissions without applying that mode to missing shared ancestors
- tighten an existing data root before FileSecretStore writes secrets, including the real config-import-first ordering
- decide from the post-`chmod` `stat(2)` result and the actual directory-entry mutation rules
- validate both the secret directory and its immediate parent against substitution
- preserve the create-at-mode path and the secrets file's existing `0600` creation and read-time checks
- reconcile `docs/linux.md` and `CONTRIBUTING.md`, regenerate the real Linux transcript, and add real-helper shell coverage

## Root cause

`AppPaths.ensureDirectories()` created the root with the process default mode, commonly `0755`. `FileSecretStore.prepareDirectories()` then returned early when that root already existed, so the documented defense-in-depth directory mode was never enforced in the normal flow.

The secrets file itself remains protected by its existing `open(O_CREAT|O_EXCL, 0600)` path.

## Resulting-mode policy

The store attempts `chmod 0700` and then inspects the resulting mode instead of trusting the syscall return value, which also covers filesystems that silently ignore permission changes.

- group/world write **and search** access is fatal when it permits entry replacement
- isolated write without search (`0720` / `0702`) is not fatal because directory entry mutation requires both bits for the same permission class
- sticky mode is honored when the directory and protected-entry ownership boundary prevents other users from replacing entries; ordinary scratch paths under `/tmp` remain usable
- group/world read or search exposure warns once per mutation and continues; it may expose directory entries and/or fixed-path file metadata, but not destination IDs stored inside the `0600` file
- the immediate parent gets the same replacement-risk check so another writer cannot rename the entire data root aside; the documented boundary does not claim to audit every ancestor or filesystem ACL

The pre-lock and in-lock checks remain, but only the in-lock store check reports the degraded-mode warning, avoiding duplicate stderr output.

## Ancestor creation

Darwin Foundation and Swift 6.1 corelibs-Foundation both apply `posixPermissions: 0700` to every directory created by a single `withIntermediateDirectories: true` call. `AppPaths.ensureDirectories()` now creates missing ancestors separately with the process-default mode, then creates only `root` with `0700`. A cross-platform test compares missing ancestor modes with an ordinary directory created under the same umask.

## Documentation transcript

I ran `scripts/linux-docs-transcript.sh` to completion on a real Linux/systemd host using the PR's x86_64 static artifact. Its actual `stat` output was `600 secrets.json` and `700 .`; the normalized transcript records that output rather than a hand-edited assumption.

## Empirical permission checks

On Linux with separate users in a shared group:

- sticky `1770`: cross-user rename denied
- non-sticky `0770`: cross-user rename allowed
- `0720` (write without search): cross-user rename denied
- `0730` (write plus search): cross-user rename allowed

These probes support the mode arithmetic used by the implementation instead of relying on the man page alone.

## Validation

The host has no Swift executable, so validation used the oldest-supported `swift:6.1` container:

- `swift test --package-path Core` — 551 tests passed
- `swift test` at the repository root — 112 tests passed
- `scripts/secret-cli-test.sh` against the Swift 6.1-built helper — passed, including config import leaving the directory `0755` before `secret set` tightens it to `0700`
- `scripts/linux-docs-transcript.sh` on the real Linux host — completed successfully
- `git diff --check` and shell syntax checks — passed

Red-checks:

- restoring the original existing-directory early return fails the ensureDirectories-then-secret-set regression because the directory remains `0755`
- removing the initial resulting-mode guard fails the `0770` refusal test because the secret is stored
- against commit `f95854c`, the review follow-up tests showed: `1777` and `0720` were wrongly refused, `0711` emitted no warning, `0755` warned twice, a non-sticky writable immediate parent was accepted, and missing ancestors were forced to `0700`
- the final Swift 6.1 suites above pass with all checks restored

Fixes #49


bherila:
Reviewed the diff and traced the parts CI cannot check. **The code change is right and the red-check is credible** — `setMode` on the existing-dir path is the correct fix, covering `EEXIST` as well is a nice catch, and moving `ensureDirectories()` to create `root` at `0700` addresses the "consider whether" half of #49. Validating in the `swift:6.1` container and saying plainly that host Swift was unavailable is exactly right.

Four things before this leaves draft. Three are documentation that this PR **contradicts**, which `CONTRIBUTING.md` requires be reconciled in the same PR.

### 1. `docs/linux.md:384-392` now states the opposite of the code

That passage argues at length that the directory "really is `0755` here … not `0700`", explains *why* (`config import` runs first and `ensureDirectories()` got there at `0755`), and links #49. After this PR both halves of that are false. This is a normative spec doc.

### 2. `docs/linux.md:429` contains a transcript that is now factually wrong

```
$ stat -c '%a %n' secrets.json .
600 secrets.json
755 .
```

That was real captured output, and this document's opening claim is that every transcript in it is real. After this change the same command prints `700 .`. Per the convention used on #51 and #52, **regenerate it from a real run rather than hand-editing the number** — `scripts/linux-docs-transcript.sh` exists for this.

### 3. `CONTRIBUTING.md:9` lists the old behaviour as a key invariant to *preserve*

> the containing directory is `0700` only when `FileSecretStore` itself creates it — it is left at `AppPaths`'s default `0755` when that directory already exists, e.g. after `config import`; see `docs/linux.md` §Secrets and issue #49

That is in the "Key invariants to preserve (each has negative tests — keep them)" list. This PR deliberately changes that invariant, which is the right call — but the list has to change with it, and the #49 reference should become the stronger guarantee.

Note for context on why CI stayed green: `scripts/secret-cli-test.sh` already asserts `DIR_MODE == "700"`, and passes both before and after, because in that script's flow `FileSecretStore` creates the directory fresh. Nothing in CI exercises the `config import`-first ordering against the documented mode, which is precisely the gap #49 was about.

### 4. Design question: `chmod` failure is now fatal, and I don't think it should be

`setMode` throws, so `prepareDirectories()` throws, so `secret set` fails outright. Previously this was best-effort (`_ = chmod(...)`). On a data directory where `chmod` cannot succeed — an SMB/exFAT mount, or a directory owned by another user — a host goes from "secret stored, file `0600`, directory looser than we'd like" to **"cannot store secrets at all"**, i.e. cannot back up. Since the file mode is the protection that actually matters and is independently re-verified on every read, making the defence-in-depth layer fatal trades a working capability for hardening.

Suggested rule, which I think is better than both the old and new behaviour — **distinguish the two loosenesses**:

- **Group/world-*writable* directory → fatal.** This is a real integrity vulnerability: another user can `unlink`/`rename` and replace `secrets.json` regardless of its `0600` mode, and the mode check on read would not notice a substituted file. Refusing is correct here.
- **Group/world-*readable* (but not writable) → warn, don't throw.** The exposure is metadata only — which destination IDs have secrets, and the file's size and mtime. That is what `docs/linux.md` already says. Breaking backups over it is disproportionate.

So: attempt the `chmod`, then `stat` and decide on the resulting mode, rather than on whether the `chmod` syscall returned 0. That also makes the behaviour correct on filesystems where `chmod` silently no-ops.

If you'd rather keep it strictly fatal, that's a defensible call — but then say so in the doc-comment, because it is a new way for `secret set` to fail and someone will hit it on a network-mounted state directory.

bherila:
Read `f95854c`. The docs reconciliation is right, the transcript is a real regeneration, and the CONTRIBUTING invariant now states the stronger guarantee. The writable-vs-readable split is implemented the way it was specified — deciding from `stat`, not from `chmod(2)`'s return — and I confirmed `prepareDirectories()` is reached only from `withWriteLock` and `store`, so `print-password` never emits the warning and `RESTIC_PASSWORD_COMMAND` output stays clean. That was the thing I was most worried about in the design I asked for.

Three findings, one of which I think is a real regression.

### 1. The sticky bit makes the refusal a false positive — and makes its stated reason untrue

`setMode` refuses on `permissions & 0o022 != 0`, and the error says *"Another user could replace secrets.json even though the file is mode 0600."* For a **sticky** directory that sentence is false. `S_ISVTX` restricts `unlink(2)`/`rename(2)` of entries to the entry's owner, the directory's owner, or root — which is exactly the substitution attack the check exists to stop. A `1770`/`1777` directory is group/world-writable and *not* substitutable.

So this refuses a configuration that is actually safe, with a reason that does not hold, and it refuses it **fatally**: `secret set` now fails outright where it previously succeeded. `RESTIC_STATION_DATA_DIR` pointed somewhere under a sticky shared directory is not exotic — a scratch dir under `/tmp` is the obvious one, and CI and throwaway hosts do that.

Suggested: treat `S_ISVTX` as satisfying the write check, i.e. only refuse when `permissions & 0o022 != 0 && (info.st_mode & S_IFMT-masked mode) & S_ISVTX == 0`. If you'd rather keep the blanket refusal, then the error message has to stop claiming substitution is possible, and the doc-comment has to say the check is deliberately stricter than the threat — but I think honouring the sticky bit is the better answer, because the current behaviour trades a working capability for a hardening measure that the kernel is already providing.

Either way this needs a test with a `1777` directory, since that is the case that decides which behaviour is intended.

### 2. The warning fires twice per `secret set`

`withWriteLock` calls `prepareDirectories()`, then `store()` calls it again inside the lock — so a readable-only directory prints the whole four-line warning twice for one command. Both call sites are deliberate (the second is what makes `store` safe to call on its own), so don't remove either; just make the warning fire once per store instance, or hoist it so the second call is silent. A duplicated warning reads like two distinct problems, and this is a message we want people to actually read.

### 3. The parent directory is unchecked, which is the same hole one level up

The check covers `root` itself. If `root`'s **parent** is group/world-writable and non-sticky, another user can `rename(2)` `root` aside and put their own `0700` directory in its place — the substitution attack, one level up, and every check here passes. I am not asking you to walk the whole ancestor chain (that ends in an argument about `/home` and `/tmp`), but the immediate parent is cheap to `stat` and covers the realistic case of a data dir inside a shared project directory. At minimum, say in the `prepareDirectories()` doc-comment that the guarantee stops at `root` and why — an undocumented boundary on a security check is how the next person builds on it wrongly.

### Not blocking, worth noting

`AppPaths.ensureDirectories()` passes `0o700` with `withIntermediateDirectories: true`. If a missing *intermediate* also picks up `0700` (e.g. a fresh `~/.local/state`), that is harmless for the owning user but it is a mode we set on a directory that isn't ours. Worth one line in the doc-comment saying which directory the attribute is intended for.

Item 1 is the one I'd call a blocker. 2 and 3 are small.

bherila:
@codex review

Context for this review — please treat all of it as *claims to check*, not as given.

This PR closes #49: the state directory that holds `secrets.json` was left at `0755` whenever something else (`config import` → `AppPaths.ensureDirectories()`) created it before `FileSecretStore` got there. Fresh directories are now created `0700` and existing ones are tightened before any secret is written.

The part that most needs adversarial attention is the new failure mode, which I specified and which is therefore under-reviewed: `FileSecretStore.setMode` decides from the mode `stat(2)` reports *after* the `chmod`, not from `chmod`'s return value, because some filesystems report success without changing anything. It then splits on the result — group/world-**writable** throws, group/world-**readable** warns and continues. That split means `secret set` can now fail where it previously succeeded, so please weigh whether the refusal is correctly scoped and whether the warning path leaves a state a caller can misread as success. I have already flagged the sticky-bit case in a separate comment; please form your own view rather than deferring to mine, and look for other modes or filesystem behaviours where the mode check reaches the wrong verdict.

Specific things worth verifying rather than assuming:

- Whether the mode arithmetic is right on Linux and Darwin, including `st_mode`'s type bits and the `UInt32` conversion.
- Whether tightening an existing directory can break a legitimate co-tenant of the state directory. It holds non-secret state too (`runs/`, `state/`, `locks/`), and this change makes a directory that other tooling may read owner-only.
- Whether `prepareDirectories()` is reachable from any read path. I believe it is not — `secret set`/`secret remove` only — but if it is, the warning would be emitted on every `RESTIC_PASSWORD_COMMAND` invocation and would land in restic's own stderr.
- The `EEXIST` race branch: another process created the directory between our `stat` and our `mkdir`. Is enforcing our mode on a directory someone else just created the right call, and is the ordering against a concurrent secret write sound?
- Whether the new tests actually fail without the fix. The PR claims two red-checks; please confirm the assertions are pinned to the specific mode/behaviour and not to a process exit code that could pass for an unrelated reason. (That exact trap bit us on #51: an integration assertion asserted `status` exit 0 where the container exits 1 on scheduler grounds regardless, so the check could only ever pass on macOS.)
- Docs: `docs/linux.md` §Secrets and `CONTRIBUTING.md`'s key-invariants entry were rewritten to match. Are they now accurate, and is the regenerated `stat` transcript consistent with what the code actually produces in that flow?

Toolchain note: `swift:6.1` is the oldest supported toolchain and decides.

bherila:
Following up on my item 4 — I verified it rather than leaving it conditional, and it is broader than I wrote.

`FileManager.createDirectory(withIntermediateDirectories: true, attributes: [.posixPermissions: 0o700])` applies the mode to **every** directory it creates, not just the leaf. On macOS (Swift 6.1, `/tmp/itest/a/b/c`):

```
700  /tmp/itest
700  /tmp/itest/a
700  /tmp/itest/a/b
700  /tmp/itest/a/b/c
```

So on a host where `$XDG_STATE_HOME` doesn't exist yet, `AppPaths.ensureDirectories()` now creates `~/.local` and `~/.local/state` at `0700` as a side effect — directories shared with every other application that follows the XDG layout. It is harmless for the owning user and I am not calling it a blocker, but we should not be setting modes on directories that aren't ours as an undocumented side effect.

Two things to do with it:

1. Create `root` in two steps — `withIntermediateDirectories: true` with no attributes for the ancestors, then set the mode on `root` alone. `ensureDirectories()` is not on a hot path and this removes the surprise entirely. If you keep the single call, then say in the doc-comment that ancestors inherit `0700` and that this is accepted.
2. **Check the Linux behaviour before deciding.** I verified the above on Darwin Foundation only; swift-corelibs-foundation is a separate implementation and may apply the attribute to the leaf only. Whichever way it goes, one of the two platforms is then doing something the doc-comment doesn't describe, which is worth a test either way — `ensureDirectoriesCreatesRootOwnerOnly` currently asserts the mode of `root` and nothing about its ancestors, so neither behaviour is pinned.

That test is otherwise the right shape (`stat`-based, asserts the mode itself rather than a proxy) — the gap is only in what it doesn't cover.

bherila:
Fable safety review done — two passes in parallel, one attacking the POSIX/threat-model logic in its own worktree (it built the branch, ran the 551 Core tests, and empirically exercised the permission arithmetic), one on tests, docs, the shell script, and operational consequences. Both were explicitly told not to defer to the earlier rounds, because the writable-vs-readable split and the sticky/parent logic were specified in *my* review — on the part of this PR most likely to be wrong, the two previous reviewers were partly reviewing their own design.

They independently converged on the same two items, and neither found a code defect that blocks. Both are fixed in fc2ac90.

**Blocker (fixed): `docs/keychain-and-fda.md` §"The `0600` rules" still documented the behaviour this PR removes.** Its last bullet said an existing `AppPaths.root` "is left at whatever mode it has" and "may legitimately have been created `0755` by `AppPaths.ensureDirectories()`". Both halves are false as of this PR, and that section is where `docs/linux.md` §Secrets and `README.md` send readers for the full threat model — so the next person reasoning about secret handling would have built on a guarantee that no longer holds. `CONTRIBUTING.md`, which this PR also edits, makes that a gate rather than a follow-up. Rewritten to state the tighten, the stat-not-return-value check, the fatal-vs-warning split, the sticky exemption, the one-level parent check, and the write-path-only scope — plus the collateral nobody would guess from the code: an other-uid reader of the *non-secret* state in that directory loses access at the first `secret set`, silently. That is the intended trade #49 asked for, but it should be written down, and it belongs in the devprod0 release note.

**Should-fix (fixed): both mode messages prescribed the exact command the code had just watched fail.** Neither the refusal nor the warning is reachable unless our own `chmod(path, 0700)` ran and the following `stat` showed the bits still set, so "Fix it with: `chmod 700 <path>`" is advice already tried. On a mode-forcing mount (CIFS/FAT `dir_mode`/`dmask`, vboxsf, WSL drvfs without `metadata`) no chmod by anyone will ever change it: the operator runs the suggestion, nothing happens, and the next `secret set` prints identical text. A correct refusal whose remedy cannot work is how a tool teaches people to ignore its output — the same failure mode as the alarm nobody reads, reached by a smaller route. Both messages now name the two reachable causes: another user owns the directory (run it as that owner or root), or the filesystem does not honour permissions, in which case the data directory has to move. Pinned by asserting the messages name `RESTIC_STATION_DATA_DIR`, red-checked against the old text — both new assertions failed with the message printed verbatim, then passed. `docs/linux.md` §Secrets says the same thing for operators.

**What came back verified rather than assumed** — the parts I most wanted checked independently:

- **The write-AND-search arithmetic is correct POSIX, tested rather than reasoned.** On a real directory: write-without-search (`0200`) fails unlink, create, and rename-in with `EACCES`; search-without-write (`0100`) likewise; `0300` allows all three. Since the permission algorithm grants exactly one class per user — a group member never falls through to the `other` bits — the per-class AND in `hasUnprotectedReplacementAccess` is the right condition. Linux `may_delete` requires `MAY_WRITE|MAY_EXEC` identically.
- **The upgrade cohort is narrow and the #49 cohort keeps working.** Existing `0755` dirs are tightened and exercised end-to-end in the green `linux-integration` job. The newly-refused set is loose-mode mounts and foreign-owned group-wx directories — and on mounts whose `chmod` is a no-op the *file* mode is usually unenforced too, so `load()` already refused those at read time before this PR. This mostly moves an existing failure earlier and makes it legible. No analogue of the #51 fresh-installs-only bug: tighten-on-write targets existing hosts by construction.
- **The private-group false-positive I was worried about is moot on the refusal path.** On any chmod-honouring filesystem the owner's own `chmod 700` succeeds *before* the check, so a `0770` private-group dir is silently fixed, never refused. Refusal requires chmod to have already failed, where group membership can't be reasoned about anyway. No crying wolf here.
- **The fail-open-read / fail-closed-write asymmetry is the right split, and it was checked rather than accepted.** Refusing reads would stop backups across the entire upgrade cohort at the next tick — a worse failure than the exposure it prevents. And a hostile directory cannot feed the read path a file that is simultaneously readable by the victim and free of group/other bits: non-root cannot `chown`, `O_NOFOLLOW` kills symlinks, ACL grants surface in the group-class mask, and hard-link substitution needs `fs.protected_hardlinks=0`. Residual read-path exposures are DoS-shaped and loud, never false assurance.
- **TOCTOU is bounded.** `store()` re-runs the full check inside the flock immediately before `unlink`/`open(O_EXCL,0600)`/`fsync`/`rename`, and every actor who could mutate the directory in that gap is one the checks already refuse, or is owner/root. Residual races degrade to visible errors, never a silent wide-mode write.
- **The tests pin the specific behaviours.** The no-op `directoryModeSetter` injections are the right tool, not an evasion — they simulate the chmod-ineffective filesystem, which is unreachable on a real POSIX test host — and the real-chmod path is separately pinned by `tightensDataDirectoryCreatedBeforeSecretSet`, both parent tests, and `secret-cli-test.sh` end-to-end. The two refusal messages' substring assertions cannot satisfy each other. Every test uses a unique-UUID temp root, so no `defer` can make a later test pass. `#require(messages.count == 1)` genuinely pins one-warning-per-mutation. `ensureDirectoriesDoesNotApplyRootModeToMissingAncestors` discriminates under CI's umask 022 and goes vacuous under 077 — where the asserted behaviour is correct by definition.
- **`secret-cli-test.sh` reproduces the #49 ordering and is red on revert**, and the new lines are bash-3.2-safe, so they hold on both the BSD `macos` and GNU `linux` jobs.

**Filed rather than fixed here**, both loud-failure notes that predate this PR:

- #62 — reads check the file mode but never `st_uid`. Self-correcting for a non-root helper (an attacker can't grant write access without setting group/other bits, which trip this PR's fatal check), but a **root-run** helper in an accepted sticky directory could `load()` an attacker-pre-created `0600` file. Pathological config, and root is documented out of scope; cheap to close.
- #63 — in a `1777` data directory a squatted `secrets.json.tmp` cannot be unlinked (`EPERM` under sticky, return value discarded), so every write dies permanently with `File exists` and nothing names the cause.

Verdict from both passes: merge. Waiting on CI for fc2ac90.

Issue #91, opened 2026-08-19
Key the config migration backup to the version being migrated from


bherila:
## The bug

`ConfigStore`'s migration copies the pre-migration bytes to a **fixed** `config.v1.backup.json`, written with `O_EXCL` so it is never overwritten. That was correct while v2 was the only migration destination. A third schema version makes it wrong.

A host that migrated v1→v2 long ago already has `config.v1.backup.json` on disk. When that host later migrates v2→v3:

1. `writeBackupIfAbsent` finds the file present and returns early — no throw.
2. `migrateToCurrentVersion` therefore reports `backupWritten: true`.
3. `load()` sees that and overwrites `config.json` with the migrated content.

The v2 file is gone and the only backup on disk is a v1 file from years earlier. This silently breaks the invariant `docs/data-model.md` §Versioning states in bold — *"the v1 file is never overwritten unless a backup of it exists"* — and it breaks it precisely for the hosts that have been running longest.

Nothing is lost today because v2 is current and the v1→v2 path writes the correctly-named file. It becomes live the moment `AppConfig.currentVersion` goes to 3, which #86 does.

## The fix

Already implemented on branch `agent/purging-exclusions`, commit `8b9158f`, while adding the v2→v3 migration:

- `AppPaths.configBackupFile(fromVersion: Int) -> URL` → `config.v<N>.backup.json`, keyed by the version being migrated **from**. `configV1BackupFile` is retained — it names a file that exists on every already-migrated host — and now delegates to it.
- `ConfigStore.writeBackupIfAbsent(_:fromVersion:)` takes the source version and uses that path.
- Tests: `aV2ToV3MigrationWritesItsOwnBackupEvenWhenAV1BackupAlreadyExists`, `migrationIsIdempotentAndNeverOverwritesAnExistingV2Backup`, and a backup-write-failure case that leaves `config.json` alone. The failure case is reproduced without `chmod`, which the pre-existing comment in that suite notes is defeated by a CI container running as root — it pre-occupies `locks/` with a plain file so `paths.ensureDirectories()` fails deterministically for any user.

Note the migration is a **single jump**, not a per-step chain: `migrateToCurrentVersion` sets `version = currentVersion` directly, so exactly one backup file is written per migration event, named for wherever the config started. A v1 host upgrading straight to a v3 build gets `config.v1.backup.json` and no v2 backup, which is correct — there never was a v2 file on that host.

## What is left — the reason this issue exists

[...]

Issue #86, opened 2026-08-19
Purging exclusions 1/5: add purgeExcludes to the config model and apply it to new backups


bherila:
Part 1 of 5 of the purging-exclusions epic (#85). Read #85 first — it carries the semantics and the six design decisions this series implements.

This task adds the config field and the **forward-only** half of the behaviour. It changes no repository data and invokes no new restic subcommand. Nothing here is destructive.

### `Core/Sources/ResticStationCore/Config/Models.swift`

- `BackupSet.purgeExcludes: [String]` — non-optional, defaults `[]`, decoded with `decodeIfPresent(…) ?? []`, and **encoded always** (matching `excludes`, not the `encodeIfPresent` exception carved out for `onboardingCompleted` and `machines`).
- `BackupSet.effectiveBackupExcludes: [String]` — `excludes + purgeExcludes`, order-stable and deduped. This is what reaches restic; nothing else should concatenate the two lists.
- `AppConfig.currentVersion = 3`, plus the changelog line in its doc comment (`- 3: purgeExcludes on BackupSet — patterns excluded going forward and rewritten out of existing snapshots.`).
- `validate()` — reject empty-string patterns in `purgeExcludes`. `excludes` is unvalidated today and the App happily saves blank rows; do not repeat that for a list that deletes data. A new `ConfigError` case with an inline-mappable message.

### `Core/Sources/ResticStationCore/Config/ConfigStore.swift`

v2→v3 step in `migrateToCurrentVersion(_:originalBytes:)`, mirrored in `previewMigration(_:)`. It is a pure default-fill — every set gains `purgeExcludes: []`. Preserve the existing ordering invariant: the pre-migration file is never overwritten unless a backup of it exists.

### Call sites

- `Core/Sources/ResticStationCore/Config/Resolution.swift` — `purgeExcludes` passes through both scopes unchanged, exactly as `excludes` does. No machine override, matching the existing design where `excludes` is not overridable either.
- `Core/Sources/ResticStationCore/Engine/BackupEngine.swift:227` — pass `set.effectiveBackupExcludes` instead of `set.excludes`. `ResticCommand.backup` keeps its single `excludes:` parameter; the caller supplies the combined list.
- `Core/Sources/ResticStationCore/Config/ConfigDiff.swift` — add `purgeExcludes` to the private `SetProjection` (a change to it must kick a tick) and to `summarize`'s per-set changed-field list, in declaration order.
[...]

Issue #89, opened 2026-08-19
Purging exclusions 4/5: standalone prune and the Maintenance UI


bherila:
> **Before starting, read [#85 §Working agreement](https://github.com/bherila/restic-station/issues/85).** It carries the repo conventions, the `--json` envelope contract (new since this issue was written), what `local-ci.sh` does *not* cover, and the review process. Everything below assumes it.

> **Envelope delta for this issue.** Any new `--json` surface follows #85 §Working agreement. The UI half is SwiftUI in `App/` and is not covered by `headless-cli-test.sh` — `local-ci.sh` builds the app but asserts nothing about its behaviour.

Part 4 of 5 of the purging-exclusions epic (#85). Depends on 3/5. Read #85 first.

`rewrite --forget` unlinks data but does not reclaim disk. Only `prune` does. Today prune exists only as `forget --prune`, so **a set with no `RetentionPolicy` has no way to reclaim space at all** — which is exactly the state a set that purges but does not forget will be in.

### Engine

`runPruneRepository(set:destination:dryRun:)` — a prune that does not require a `RetentionPolicy`. Keeps `runPrune`'s mirror-staleness guard: never prune a mirror that is behind the primary. Set lock, secret preflight, reachability probe as usual.

### Helper

`restic-station maintenance prune --set <uuid> [--dest <uuid>] [--dry-run] [--json]`.

### App

- Extend `App/Sources/Views/Maintenance/RetentionSection.swift` with a "Reclaim space" action distinct from "Apply retention", and a purge section showing the pending purge plan with its warning.
- Wire through `AppModel+Maintenance.swift`, which already has the dry-run-then-confirm pattern used by `previewCleanup` / `prepareApplyRetention` — reuse it rather than adding a second confirmation flow.
- The iCloud warning already shown for repos under `~/Library/Mobile Documents` should be repeated on the prune action. For a repository on iCloud Drive, prune means downloading and re-uploading pack files, and the repository must be fully materialised locally — not left as evicted placeholders — or restic will stall or fail partway.

## Docs

`docs/ui-spec.md` for the new Maintenance controls; `docs/restic-cli.md` §prune already lands in 2/5.

## Acceptance criteria

- [ ] A set with `retention == nil` can be pruned.
- [ ] A mirror behind the primary is never pruned, and says why.
- [ ] `--dry-run` spawns no command that can modify the repository.
- [ ] The prune action warns before starting on a repository under `~/Library/Mobile Documents`.
[...]

PR #92, opened 2026-08-19
Complete purge exclusion surfaces

bherila:
## Summary

Completes the remaining work for #86 and #91 on top of the existing schema-v3 and migration commits.

- Adds the distinct purge-excludes editor and effective-config JSON/human-report fields.
- Documents the dangerous history-facing semantics and prune requirement, and updates current fixtures/transcripts to schema v3.
- Corrects version-keyed migration-backup messages/comments/docs and adds coverage for source-versioned paths.

## Why

Purge exclusions must stay separate from ordinary backup exclusions: they prevent future backups and carry the intent to remove matching paths from existing snapshots. Migration backups must be named for the schema version actually being replaced so a v2-to-v3 upgrade cannot overwrite the only recoverable copy.

## Validation

- `swift test`
- `swift test --package-path Core`
- `./scripts/local-ci.sh`


bherila:
Reviewed the branch and pushed three follow-up commits. Two of the three were things neither `local-ci.sh` nor CI could have surfaced on its own.

### `cc1a67a` — CI was red, not green

The `linux` job failed on **"restic discovery resolves the helper's binary on Linux"**. That step asserts a v1 `config.json` migrates in place and still required `"version": 2`.

It is the **only** migration assertion that lives in `.github/workflows/ci.yml` rather than under `scripts/`, which is exactly why `local-ci.sh` reported green while CI did not. I bumped the literal and left a comment at the site saying so, so the next schema bump has a chance of finding it.

The neighbouring `config.v1.backup.json` assertion is unchanged and still correct — migration is a single jump, so a v1 file yields a v1-named backup regardless of which version it lands on. Verified locally against a real v1 file: `config.json` at 3, `config.v1.backup.json` at 1, no v2 backup written.

### `c7c4e8d` — four normative transcripts went stale silently

`EffectiveConfigReport.humanLines()` gained `excludes:` and `purge excludes:` lines. `docs/linux.md` quotes that function's **real output** in four `config validate` transcripts, so all four were two lines short per set.

Nothing can catch this: CI *produces* those transcripts (the "Docs transcript for docs/linux.md" step) but never diffs them against the doc, and `CONTRIBUTING.md` treats the drift as blocking. I regenerated each block against the fixture it comes from — `config-v2.json` for the two `--machine` examples, the inline mirror-box JSON, and the rendered `mac-exported-config.json` — and diffed; all four now match byte for byte, including the `node_modules, .build, *.tmp` excludes the fixture actually carries.

Same commit: `config import`'s abstract and its `--dry-run` note still promised a "v1→v2" migration that now targets v3. Both are user-facing strings, and one of them is quoted in `docs/linux.md`. Reworded to name no version at all, along with three comments, so a future bump cannot stale them again.

### `0c90e0d` — a copy string described the wrong operation

The purge-excludes empty state read *"No existing snapshots are marked for removal."* That describes deleting snapshots. Purging does the opposite: `rewrite --forget` replaces each snapshot with one that omits the matched paths, and the **snapshot count is unchanged** — the property #89's integration test is meant to assert. Reworded in `SetsCopy` and in the ui-spec entry that pins it verbatim.

### What I checked and found correct

The rest of `9014273` holds up. Specifically worth noting:

- The `config import` backup-failure message now names the source-versioned file *and* the source version, and `ConfigImportTests` covers it — that closes #91's one functional item.
- `humanLines()` printing excludes at all is a genuine pre-existing gap fixed, not just a new field bolted on.
- The `ui-spec.md` copy matches `SetsCopy` verbatim, which is what makes that file reviewable against the spec.
- `docs/data-model.md`'s new "single jump" paragraph is the right thing to have written down — it is the non-obvious part of the version-keyed backup, and I confirmed the behaviour matches the prose.

Local: `swift test` (77), `swift test --package-path Core` (588), `scripts/headless-cli-test.sh`, and the app build via `xcodebuild` after `./scripts/bootstrap.sh` are all green.

bherila:
@codex review

Issue #82, opened 2026-08-16
Agent-ready CLI 5/7: require expiring preview tokens for destructive retention


bherila:
## Context

Retention is Restic Station's destructive operation: forget can remove snapshot metadata and prune can delete repository data. A generic --yes flag proves only that a caller supplied a boolean; it does not prove that the caller reviewed the current policy against the current repository state.

This is item 5 of the agent-ready CLI series. Any implementation PR for this issue must carry the fable-review label.

## Goal

Bind retention application to a recent, exact, read-only preview. If configuration, repository snapshots, destination eligibility, or time changes, application must fail closed and require a new preview.

## Proposed flow

    restic-station retention preview --set <set-uuid> --json

The response includes a short-lived opaque previewToken plus the human/auditable plan.

    restic-station retention apply --set <set-uuid> --preview-token <token> --json

There must be no --yes, --force, environment-variable, or interactive shortcut that bypasses the token for agent-facing CLI use.

## Token requirements

The token must be:

- cryptographically random and unguessable;
- scoped to machine ID, effective set ID, exact destination IDs/roles, policy fingerprint, config fingerprint, repository identities, and the exact snapshot removal plan;
- short-lived, with a documented default such as 15 minutes;
- single-use;
- stored only in a local mode-0600 state location or represented by a securely authenticated payload;
- absent from normal logs, run records, telemetry, and error details.

A UUID alone is acceptable only if it indexes protected local state containing all bindings and sufficient entropy comes from the system CSPRNG.

## Apply-time validation

Under the set lock and before the first destructive command:

1. Resolve the effective set again.
2. Load and validate the token, expiry, machine, set, and destination scope.
3. Probe every destination included in the plan.
4. Re-run retention preview for every destination.
5. Compare policy, config, repository identity, snapshot IDs, keep/remove decisions, and destination eligibility to the token-bound plan.
6. Refuse the entire operation before pruning anything if any comparison differs.

Maintain the existing safety rule that a secondary is eligible only when it is proven synchronized with the primary. Validate all destinations before mutating the first one. Apply to eligible secondaries before the primary, consistent with the current safety posture, and record per-destination outcomes.

If a command partially succeeds because a later repository fails, return a distinct bounded result that identifies completed and unattempted destination IDs and invalidates the token. Never retry destructive work automatically.

## Token lifecycle

- Expired token: preview_expired.
- Changed config/repository/plan: preview_stale.
- Used token: preview_already_used.
[...]

Issue #79, opened 2026-08-16
Agent-ready CLI 2/7: provide uniform JSON output for inspection and probe commands


bherila:
## Context

Issue #29 added the initial headless CLI and JSON modes for status and list operations. The remaining inspection and probe commands still have uneven machine-readable coverage, forcing agents and monitoring scripts to scrape human text or discover command-specific exceptions.

This is item 2 of the agent-ready CLI series.

## Goal

Every non-destructive inspection or probe operation should offer a stable --json mode, follow one stdout/stderr convention, and use shared Codable response models rather than ad hoc dictionaries.

## Required inventory

Audit the complete command tree and document a command/output matrix. At minimum, cover:

- version --json;
- config validate --json;
- sets list --json (existing contract, verify and document);
- status --json (existing contract, verify and document);
- runs list --json (existing contract, verify and document);
- runs show <run-id> --json;
- probe-repo --set <uuid> --dest <uuid> --json;
- secret list --json, returning presence metadata only;
- cli status --json;
- the backup dry-run, snapshot-list, retention-preview, and capabilities commands from this series when present.

config export should remain the exported config document itself rather than being wrapped in an unrelated response shape. Commands that intentionally stream NDJSON should be explicitly identified as streaming contracts.

## Output rules

For every --json mode:

- stdout contains exactly one JSON document, or a documented NDJSON stream, and no progress prose, warnings, ANSI escapes, or logging;
- diagnostics that are not part of the contract go to stderr;
- success fields use consistent casing, UUID rendering, timestamps, byte counts, and outcome vocabulary;
- Optional fields use consistent null-versus-omitted behavior;
- no password, secret environment value, Keychain output, command environment, or raw authentication error is serialized;
- human output remains the default and stays readable;
- NO_COLOR and non-TTY behavior remain correct.

A probe may update its normal reachability state, but JSON support must not add new side effects.

## Compatibility and versioning

Document every JSON schema in docs/data-model.md or a dedicated docs/cli-json.md. Treat these as public automation interfaces. Add schemaVersion where a response is not already governed by config schema versioning.

Coordinate failure output with the stable JSON error-envelope issue in this series. Existing successful JSON consumers should either remain compatible or receive an explicit migration note and compatibility fixture.

## Implementation guidance

Prefer a shared output abstraction:

- typed response structs in Core;
- one encoder configuration for sorted keys, dates, UUIDs, and optional values;
- one JSON emission path that owns stdout;
[...]

PR #93, opened 2026-08-19
Add a stable versioned JSON error envelope to the helper CLI

### LINKED ISSUE #81: Agent-ready CLI 4/7: define a stable versioned JSON error envelope
## Context

Exit codes are intentionally small and useful for shell automation, but an agent also needs to distinguish set-not-found, busy, offline, uninitialized, secret-unavailable, dataless cloud repository, invalid config, and restic execution failures without scraping English prose.

This is item 4 of the agent-ready CLI series and should become the shared error contract for every --json command.

## Proposed contract

On a handled failure in JSON mode, stdout contains exactly one envelope:

    {
      "schemaVersion": 1,
      "ok": false,
      "error": {
        "code": "repository_offline",
        "message": "The destination is offline.",
        "retryable": true,
        "details": {
          "setId": "...",
          "destinationId": "..."
        }
      }
    }

The process still exits nonzero. Human mode remains unchanged.

The message is presentation text and may evolve. code, retryable, schemaVersion, and documented details keys are the stable automation contract.

## Initial error taxonomy

Define a centralized enum and mapping for at least:

- invalid_arguments;
- config_missing;
- config_invalid;
- set_not_found;
- set_disabled_here;
- destination_not_found;
- destination_disabled_here;
- set_busy;
- repository_offline;
- repository_not_initialized;
- repository_locked;
- cloud_repository_not_hydrated;
- secret_unavailable;
- restic_not_found;
- restic_unsupported;
- restic_failed;
- operation_not_allowed;
- preview_required;
- preview_expired;
- preview_stale;
- preview_already_used;
- internal_error.

Use a smaller taxonomy if evidence shows some cases cannot be distinguished reliably, but never infer a precise code from fragile English matching when the underlying typed error is available.

## Exit-code relationship

Preserve the current process contract:

- 0 success;
- 1 terminal/config/general error;
- 2 set busy;
- 3 offline where currently specified.

Document that exit codes provide coarse shell behavior while error.code provides the precise machine classification. Do not allocate a new exit code for every error.

## Output and privacy rules

- JSON envelope goes to stdout so a caller requesting --json can always parse stdout.
- Incidental diagnostics may go to stderr but must never be required to classify the error.
- Never include a password, secret environment value, RESTIC_* secret, raw Keychain output, full subprocess environment, or unbounded raw restic stderr.
- details should prefer IDs and safe enum values over repository URLs and source paths.
- Include an optional diagnosticReference such as a run ID or log path only when it is safe and useful.
- Cancellation and timeout must be distinct from internal errors.
- Unknown future fields must be safe for clients to ignore.

Argument-parser failures that occur before a command knows JSON was requested need an explicit design decision and documentation. If fully structured parse failures are deferred, test and document that boundary rather than pretending it is covered.

## Implementation guidance

Map typed domain errors to the envelope in one place. Command handlers should not hand-author JSON errors. The human renderer and JSON renderer should consume the same typed failure.

Add public documentation with:

- complete code table;
- retryability semantics;
- exit-code mapping;
- details fields by code;
- compatibility/versioning policy;
- redaction policy.

## Tests

- One test per error code from an injected typed failure.
- Stable golden fixtures for representative errors.
- stdout always parses as one JSON document on handled failure.
- Correct process exit code and retryable flag.
- Secret-canary checks across stdout and stderr.
- Unknown/internal errors are bounded and do not serialize arbitrary object descriptions.
- macOS Keychain and Linux file-secret backend failures map to the same logical code.

## Acceptance criteria

- [ ] All JSON-capable commands use the centralized envelope.
- [ ] Error codes and exit-code relationships are documented as public interfaces.
- [ ] Callers never need to regex human messages.
- [ ] Sensitive backend or restic output cannot leak through error serialization.
- [ ] Compatibility tests pin schema version 1.

bherila:
Closes #81.

## The problem

The helper's five `--json` commands emit well-formed JSON on success and **nothing at all on stdout when they fail**. Every failure goes through `HelperExit.fail`, which writes an English sentence to stderr and exits. An agent calling `status --json` against a host with a broken `config.json` gets an empty stdout, exit 1, and prose it would have to regex.

Exit codes stay deliberately coarse (`HelperExitCode`: 0 ok, 1 error, 2 busy, 3 offline) — right for `&&` in a shell, useless for telling `set_not_found` from `repository_locked` from `secret_unavailable`.

## What lands

```console
$ restic-station-helper sets list --json      # config.json is corrupt
{
  "schemaVersion" : 1,
  "ok" : false,
  "error" : {
    "code" : "config_invalid",
    "message" : "could not load configuration: …",
    "retryable" : false
  }
}
$ echo $?
1
```

Rather than allocate an exit code per failure — which would break every existing caller — the precise classification moves into the payload and **the exit code stays exactly what it was**.

The taxonomy, its `HelperExitCode` mapping and the envelope live in **Core**, next to the exit-code enum they map onto: splitting "which error" from "which exit code" across two packages would leave two half-contracts, and the mapping switches over Core's own typed errors, so Core's suite exercises every code without spawning a process.

`docs/cli-json.md` is the normative contract — code table, retryability semantics, `details` fields, redaction, versioning, and the two boundaries it does not cover.

## Decisions worth reviewing

**Two codes the issue's taxonomy could not carry.** Both surfaced from writing the tests, not the code:

- `secret_rejected`, split from `secret_unavailable`. restic exit 12 means the secret was read fine and *refused*. Folding them forces one `retryable` value that is wrong for one of them — a locked keychain is worth retrying unchanged, a wrong password never is.
- `run_not_found`. `runs show <unknown-id>` is a real failure of a `--json` command and had no code at all.

**Four listed codes are deliberately not defined**, each recorded in the doc: `config_missing` (a missing `config.json` is an empty config — `ConfigStore.load()` returns `AppConfig()`), the `preview_*` family (#82/#88 land the mechanism), and `cloud_repository_not_hydrated` (no detector exists; inferring it would mean matching restic's English stderr — the exact practice this contract exists to stop). A code nothing can emit invites callers to branch on a case that never arrives.

**`main()` is hand-written**, reproducing the one `AsyncParsableCommand` synthesizes plus two catch clauses. That is what makes an argument-parser failure classifiable at all — `run()` is never reached for one. Everything it does not own goes straight back to `exit(withError:)`, which keeps usage text, `--help` exiting 0, and `EX_USAGE` intact.

**`invalid_arguments` can exit 64 or 1.** A parser failure keeps ArgumentParser's `EX_USAGE`; a command validating its own arguments (`runs list --limit 0`) exits 1 as it always has. The envelope describes the exit contract rather than redefining it, so both modes always agree.

**Redaction is enforced by shape.** `CLIErrorDetails` is a fixed struct with nowhere to put a repo URL, source path, password, or stderr blob — not a rule each call site must remember. `message` is capped at 500 chars; before it, a bad `config.json` printed the whole `DecodingError` description including the bytes it choked on.

## Verification

`swift test` (86), `swift test --package-path Core` (608), and `./scripts/local-ci.sh` — all 8 stages PASS.

**Human mode is byte-identical**, verified by diffing `--help` and a representative failure against the pre-change binary. That check earned its keep: handing `CLIFailure` to `exit(withError:)` printed the raw struct instead of the sentence, and the diff caught it.

**The doc's code table is checked by tests** — every code has a row, documented retryability and exit codes match the implementation, and no row names a code that does not exist. I confirmed all three actually fail on drift rather than passing vacuously.

Three assumptions of mine were wrong and the tests caught each:

- `runs list` never loads `config.json`, so a broken config is correctly *not* a failure for it. Excluded from that loop with a comment; its own failure path is asserted separately.
- The restic-not-found branch is unreachable on any Mac with Homebrew restic — my first attempt to check it by hand exercised nothing. Now pinned by a test.
- `integration-test.sh` matched the run-index path as a literal substring, and `JSONEncoder` escapes forward slashes (`\/var\/folders\/…`). The contract is fine — `jq -r` decodes it — but the assertion was reading raw bytes. It now decodes with `jq` and asserts the classification too.

I also swept `.github/workflows/` and `linux-docs-transcript.sh` directly for anything reading a failing `--json` command, since that class is what took PR #92 red. One exists (`status refuses to report healthy when it cannot read the run history`); I reproduced its exact condition locally and its grep still matches inside `message`. Every other `--json` call there is a success path.

## Not in scope

The other 11 subcommands stay human-only and still write prose to stderr; #79 converts them as it gives them `--json`. That boundary is stated in the doc rather than papered over. `repository_offline` and `operation_not_allowed` are defined but have no producer yet, and their absence is asserted rather than assumed.


bherila:
@codex review

bherila:
Codex's four P2 findings were all real; fixed forward in 979b44c.

- **`itemNotFound` published as retryable** — new `secret_not_configured` code (not retryable). The "both backends map to the same logical code" criterion still holds: both report a missing item as `itemNotFound` and everything else as `backendFailed`, so the split is by condition, not by backend.
- **Typed `ConfigError` lost the `could not load configuration:` prefix** — now classified for its code/details, then prefixed, restoring the pre-#81 human-mode wording and naming which of the two files failed.
- **The 500-char cap depended on constructors remembering it** — bounding moved into `CLIFailure.init`, which is what its own doc comment already claimed.
- **`versionFound` copied a probed binary's stdout verbatim** — now the dotted numeric triple the too-old decision was actually made on; the raw text still reaches a human through the capped `message`.

Each is pinned by a test that fails without the fix, and `docs/cli-json.md` carries the new code, the widened retryability rationale, and the sanitization rule. `local-ci.sh` green (8/8).

bherila:
@codex review

bherila:
Second round fixed forward in 1192970 — the finding was right, and it was the more important half of the first one.

Splitting `secret_not_configured` out in the classifier accomplished nothing on the path that actually runs restic: `ResticRunner`'s pre-flight caught every `SecretStoreError` and rethrew one data-free `secretsUnavailable`, so `itemNotFound` was erased before anything could classify it. New `ResticRunnerError.secretsNotConfigured(destinationId:)`, produced by both the password and secret-env paths, still discarding the backend's own text.

Its category is `terminal`. Nothing but the CLI classifier reads `ResticRunnerError.category`, so no engine behaviour changes — but `details.resticCategory` is published next to `retryable`, and would otherwise have said "retryable" beside `false` in the same envelope.

`Reachability` had to decide too. It reports `.offline` with a reason of its own rather than the store-unreadable one: `.error` would have been truer but carries a `ResticExitClass`, and restic never ran — inventing `wrongPassword` would publish an exit code that did not happen. The reason string sits deliberately outside `SetsBadges`'s environmental list, so the badge reads "Error" rather than "Offline — try later".

`architecture.md` §Error taxonomy and `keychain-and-fda.md` both said the pre-flight failure is retryable full stop; both now name the exception. The Reachability assertion was confirmed to fail against a pre-flight that collapses the two cases. `local-ci.sh` green (8/8).

bherila:
@codex review

bherila:
Third round in e717b51. One fixed, one deliberately not.

**Fixed — the cap produced 501-character messages.** `bounded(_:)` took the first 500 characters and then appended the ellipsis, so every truncated message was one over the limit the type publishes and `cli-json.md` documents; a caller enforcing 500 would reject exactly the envelopes the cap exists to make acceptable. The ellipsis is now inside the budget, and the assertions that expected `limit + 1` expect `limit`.

**Not fixed here — the engine pre-flights.** The finding is right, and the previous commit's taxonomy table overstated it: `BackupEngine.runSet`'s step 1 and `secretsAvailable(for:)` do still collapse every secret failure into the retryable row, so a destination whose password was never stored is skipped silently on every tick with nothing recorded.

Changing that is a product decision, not a classification fix — `.failed` run every tick, `.misconfigured`, or a health warning that is not a run at all, each with different `HealthDerivation` and badge consequences. Filed as #95. The table now says what the code does and names the gap instead of papering over it; the classification and the envelope's advice are unchanged and still correct.

`local-ci.sh` green (8/8).

bherila:
@codex review

bherila:
Fourth round fixed forward in 93164fa. Right again, and it is the same mistake one level down: the fixed struct bounds *what* can appear in `details`, and says nothing about *how much*. `machineId` comes from `config.json`, `MachineIdentity` imposes no length limit, and both disabled-here factories copy it in unchanged — so "bounded by construction" was true of the key set and false of the document a caller receives.

Every free-form value is now capped at 128 characters, applied in `encode(to:)` rather than in the factories: the fields are `public var`, so a cap on the way in can be undone by assignment, while one on the way out holds for every value however it arrived. Truncation keeps a trailing `…`, so a shortened id reads as shortened rather than as a different id that might exist. `setId`/`destinationId` are `UUID` and `resticExitCode`/`resticCategory` an integer and a closed enum, so they need nothing.

Both assertions were confirmed to fail against the unbounded encoder — one on the value length, one on the whole envelope's byte count (3746 bytes for a single oversized machine id). `local-ci.sh` green (8/8).

bherila:
@codex review

bherila:
Fifth round: one fixed in 2e7b4b3, one filed as #96.

**Fixed — timeouts.** `ResticRunnerError.timedOut` classified as `restic_failed` (`retryable: false`) while the same error's `userFacingMessage` has always ended "then try again". One failure, two halves, opposite advice — and the machine-readable half was the wrong one. New `operation_timed_out`, retryable, exit 1 (unchanged; the envelope describes the exit contract rather than redefining it).

`details.resticCategory` is deliberately omitted for it, and this is the one case where the two axes genuinely disagree: the category answers "does this write a `.failed` run record" — for a timeout, yes, the operation did not complete — while `retryable` answers "can the caller repeat the request" — also yes. Publishing `terminal` beside `retryable: true` would read as advice it is not, so the field is absent and `cli-json.md`'s `details` table says why.

**Filed — `backendFailed`.** The finding is correct: the file backend raises it for permanent refusals (symlinked or group-accessible `secrets.json`, untrusted owner, malformed contents, newer format) as well as for transient failures, so all of them publish `retryable: true`.

Not fixed here because the safe direction is asymmetric. Marking a genuinely transient failure permanent breaks the pre-login-tick retry `keychain-and-fda.md` §2 depends on, and the errno cases are not obvious — `EACCES` on open can be Full Disk Access not yet granted, which does resolve. That is ~20 throw sites needing individual judgement in the secret store, with 24 test references pinning the current shape. #96 enumerates the six read-path sites and the suggested `storeUnusable` split; the code table now names the gap instead of implying it isn't there.

`local-ci.sh` green (8/8).

PR #94, opened 2026-08-19
Uniform JSON output for every inspection and probe command

### LINKED ISSUE #79: Agent-ready CLI 2/7: provide uniform JSON output for inspection and probe commands

bherila:
Closes #79.

> **Stacked on #93 (issue #81).** Base is `agent/cli-error-envelope`, not `main` — this builds directly on the error envelope. Merge #93 first, then this retargets to `main` as a plain rebase.

## What changes

Every `--json` command now emits one document with the same three top-level keys, so a caller reads `ok` once and knows which key holds the payload — without knowing which command it called.

```json
{ "schemaVersion": 1, "ok": true,  "data":  { … } }
{ "schemaVersion": 1, "ok": false, "error": { … } }
```

The wrapping lives in `CLIJSON.print`, the one path that owns stdout, so the five commands that already had `--json` wrapped without per-command edits and **no command can opt out by accident**. `config export` is the single deliberate exception — it emits the config document itself, unwrapped, so it round-trips into `config import`.

`--json` added to six more commands, for **eleven** total:

| | payload |
|---|---|
| `version` | `{ name, version, platform }` |
| `config validate` | `{ machineId, errors, warnings, effective, nothingRunsHere }` |
| `probe-repo` | `{ setId, destinationId, label, outcome, reachable, reason }` |
| `secret list` | `[{ destId, label, setName, hasPassword, secretEnvCount }]` |
| `cli status` | `CLIInstaller.Status` |
| `fda-check` | `{ applicable, granted, probedPath, checkedAt, context }` |

`docs/cli-json.md` carries the full matrix, including **every command that does not have `--json` and why**.

## ⚠️ Breaking change

`status --json` was the `StatusReport` object itself; `sets list --json` was a bare array. Both are now under `data`.

```console
# before                                   # after
… status --json | jq -r '.health'          … | jq -r '.data.health'
… sets list --json | jq 'length'           … | jq '.data | length'
```

Chosen over an additive shape deliberately, and it was your call on the trade: two conventions coexisting — some commands wrapped, some bare, arrays unable to carry a version key at all — would have been permanent, against one `.data` per caller while the tool is pre-1.0. Migration note is in `docs/cli-json.md` §Migrating from the unwrapped shape.

Every in-repo consumer is migrated: both shell suites, the transcript script, `ci.yml`, and three `docs/linux.md` transcripts. Deliberately **not** migrated: `jq` against `config.json`, `machine.json`, `state/fda-check.json`, and restic's own `--json` — those are files and other programs' output, and a blanket rewrite would have broken them.

## Decisions worth reviewing

**`probe-repo` reports offline as a success** — `outcome: "offline"`, `ok: true`, exit 3. An unplugged drive is a destination's expected state, not a fault; an error envelope would make a sleeping NAS indistinguishable from a broken config. Branch on `outcome`.

**`secret list` can have a JSON mode at all** because its row type carries *counts*, never values — the same redaction-by-shape rule `CLIErrorDetails` follows. Asserted structurally in the shell suite, not just by the end-of-run secret grep.

**`config validate` reports an unloadable config as the shared `config_invalid` error envelope**, not as a success document with a populated `errors` list, so a caller needs no per-command exception. Human mode keeps its own `Errors:` block byte-for-byte.

**`timer status` stays human-only**, with a documented reason rather than a shrug: its report is narrative assembled while probing, and its machine-readable equivalent already exists as `status --json`'s `.scheduler` — in the same problem vocabulary (`unitsMissing`, `agentNotLoaded`) that CI already asserts on.

## Verification

`swift test` (89), `swift test --package-path Core` (608), and `./scripts/local-ci.sh` — all 8 stages PASS.

New `headless-cli-test.sh` §9 runs **all eleven** commands against a real binary and asserts stdout is exactly one envelope with `data,ok,schemaVersion` **and no other top-level key** — a command that escaped `CLIJSON` shows up there. The Swift suite pins the same eleven through `JSONRenderable`, so a new `--json` command that forgets the conformance fails as a missing entry rather than as silent prose on stdout.

Two things testing turned up that reasoning had not:

- **The new payloads omitted nil fields.** Swift's synthesized encoder uses `encodeIfPresent`, but every existing `--json` shape emits explicit `null` and `data-model.md` states that as a convention. `probe-repo`, `fda-check` and `cli status` now have hand-written encoders — a missing key and a null one must not become two ways of saying the same thing.
- **My first `probe-repo` assertion pinned `"offline"`**, which was really asserting that the fixture's repository path did not exist. It now pins the whole mapping instead: every outcome is a success envelope, and each implies its exit code.

The three `docs/linux.md` transcripts were rewritten by hand and then re-parsed to prove each is a valid envelope with exactly those three keys — CI produces those blocks but never diffs them against the doc, the same gap that took #92 red.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,7 +174,12 @@ jobs:
           echo "$OUT"
           echo "$OUT" | grep -q "discovered $FAKEBIN/restic"
           # Got past restic resolution and on to the real (absent) set.
-          echo "$OUT" | grep -q "no backup set with id"
+          # Case-insensitive on purpose: this asserts *where* execution got
+          # to, not the sentence. The wording is presentation text that moves
+          # as commands convert to `CLIFailure` (#81) — probe-repo's went from
+          # a lowercase `HelperExit.fail` fragment to a capitalised sentence
+          # and turned this step red for nothing but the leading letter.
+          echo "$OUT" | grep -qi "backup set with id"
 
           # 3. tick still exits 0 on a host with nothing configured.
           #    (tick resolves restic through the same `resolveResticPath`
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,9 +230,9 @@ jobs:
             echo "$OUT"; echo "status exited non-zero on a host with no systemd"; exit 1
           }
           echo "$OUT"
-          echo "$OUT" | jq -e '.scheduler.kind == "unknown"' > /dev/null
-          echo "$OUT" | jq -e '.scheduler.healthy == null' > /dev/null
-          echo "$OUT" | jq -e '.health == "idle"' > /dev/null
+          echo "$OUT" | jq -e '.data | .scheduler.kind == "unknown"' > /dev/null
+          echo "$OUT" | jq -e '.data | .scheduler.healthy == null' > /dev/null
+          echo "$OUT" | jq -e '.data | .health == "idle"' > /dev/null
 
       - name: status refuses to report healthy when it cannot read the run history
         run: |
--- a/Core/Sources/ResticStationCore/Support/CLIError.swift
+++ b/Core/Sources/ResticStationCore/Support/CLIError.swift
@@ -725,6 +725,45 @@ extension CLIFailure {
     }
 }
 
+// MARK: - CLISuccessEnvelope
+
+/// The exact document a `--json` command writes to stdout when it succeeds
+/// (`docs/cli-json.md`, issue #79).
+///
+/// ```json
+/// { "schemaVersion": 1, "ok": true, "data": { … } }
+/// ```
+///
+/// Deliberately the same three top-level keys as ``CLIErrorEnvelope``, with
+/// `data` where that has `error`. A caller reads `ok` once and knows which
+/// key to look in — it never has to probe for the presence of `error`, and
+/// never has to know which command it called to know the shape.
+///
+/// **This replaced a bare payload**, so it is a breaking change for anyone
+/// who was reading `.health` rather than `.data.health`. The migration note
+/// is in `docs/cli-json.md` §Migrating from the unwrapped shape.
+public struct CLISuccessEnvelope<Payload: Encodable>: Encodable {
+
+    /// Shared with ``CLIErrorEnvelope/schemaVersion`` on purpose: the two
+    /// are one contract with two branches, and a caller that has pinned
+    /// `schemaVersion: 1` has pinned both. Bumping one without the other
+    /// would mean a document whose version says nothing about half the
+    /// shapes it can take.
+    public static var schemaVersion: Int { CLIErrorEnvelope.schemaVersion }
+
+    public let schemaVersion: Int
+    /// Always `true`. Present so `ok` is the single discriminator rather
+    /// than "does an `error` key exist".
+    public let ok: Bool
+    public let data: Payload
+
+    public init(_ data: Payload) {
+        self.schemaVersion = Self.schemaVersion
+        self.ok = true
+        self.data = data
+    }
+}
+
 // MARK: - CLIErrorEnvelope
 
 /// The exact document a `--json` command writes to stdout when it fails.
--- a/Core/Sources/ResticStationCore/Support/CLIError.swift
+++ b/Core/Sources/ResticStationCore/Support/CLIError.swift
@@ -518,10 +518,15 @@ extension CLIFailure {
     /// The path stays in `message` and out of `details` — a data-directory
     /// path carries a home directory, and `details` is the half of the
     /// envelope that promises to be safe to log.
-    public static func machineIdentityUnreadable(path: String, underlying: any Error) -> CLIFailure {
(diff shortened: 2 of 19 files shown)


bherila:
@codex review

bherila:
All three fixed forward in 09def16. Both P1s are the same shape — a command advertises `--json`, and a failure *before* its body runs exits directly, so stdout is empty and a caller committed to `jq` gets nothing. Exactly what #81 exists to remove, in two places #81 did not reach.

**`secret list --json`.** Its setup is `SecretContext.make()`, not `HelperContext.make()` — entering a password has to work before restic is configured, so the `secret` commands got their own path, and that path kept its `HelperExit.fail`. `SecretContext.make()` and `HelperContext.makeSecretStore` now throw `CLIFailure`; `makeTolerant` becomes `throws` to carry it, which is the same exit 1 `tick` already produced for a misconfigured backend. The backend-selection failure classifies as `config_invalid` rather than the retryable `secret_unavailable` — a bad `RESTIC_STATION_SECRET_BACKEND` will not fix itself.

**`config validate --json` without `--machine`.** The `try?` threw away the reason `machine.json` could not be read before exiting. The error is kept now; `localMachineId` stays optional because with `--machine` given an unreadable local identity genuinely is not fatal, but when it *is* the answer, why it failed is the whole failure.

**The P2 is right too, and worse than it sounds.** Human mode filters through `hasAnything`; JSON mode serialized every configured destination. An empty store answered "no destination has a stored password" to a person and a full array of `hasPassword: false` rows to a script — two modes disagreeing about the result set, from one shared payload.

All three were confirmed against the unfixed binary: 0 bytes on stdout for both P1s, and a one-row array beside the "nothing stored" sentence for the third. `headless-cli-test.sh` covers all three now, and its secret-list section stores and then removes a password so the two modes are compared with something actually in the store. `local-ci.sh` green (8/8).

bherila:
@codex review

Issue #88, opened 2026-08-19
Purging exclusions 3/5: destructive purge apply, preview token, and automatic purge on the next run


bherila:
> **Before starting, read [#85 §Working agreement](https://github.com/bherila/restic-station/issues/85).** It carries the repo conventions, the `--json` envelope contract (new since this issue was written), what `local-ci.sh` does *not* cover, and the review process. Everything below assumes it.

> **Envelope delta for this issue.** Same rules as #87 for any new `--json` surface. A token that does not match the current plan is a refusal, not an internal error: `operation_not_allowed` is defined, documented, and currently has no producer — it is the intended code. A token that has expired needs a new code (`preview_expired` and friends are reserved in `docs/cli-json.md` §Codes deliberately absent, to be defined by whichever of #82/#88 lands first); adding one is additive and must reach the code table and `CLIErrorTests`.

> **`fable-review`:** this is the task where data starts being deleted. Human review before merge, regardless of what automated review says.

Part 3 of 5 of the purging-exclusions epic (#85). Depends on 2/5. Read #85 first — design decisions 2, 3 and 4 govern this task.

**This is the destructive one.** It deletes data from backup repositories, and it makes a scheduled job capable of doing so. The implementation PR must carry the `fable-review` label.

### `Core/Sources/ResticStationCore/Support/PreviewToken.swift` (new)

A CSPRNG token indexing local state at mode `0600` under `state/`, carrying machine ID, set ID, destination IDs, config fingerprint, the exact snapshot-ID list, and an expiry (default 15 minutes). Single-use. Absent from run records, logs, and error details.

Build it to satisfy #82 as well, so that issue consumes this component rather than inventing a second mechanism. If anything about the resolved set, the destination set, the config, or the repository's snapshots differs at apply time, validation fails closed and a new preview is required.

### `BackupEngine.runPurge(set:destinations:token:)`

Modelled on the existing `forgetChild` and `runPrune`: set lock, secret preflight, per-destination reachability, then `rewrite --forget` over the token's snapshot IDs. New `RunKind.purge`; phase `"purging-<destId>"`.

The run's metadata records the old→new snapshot ID mapping. That mapping is the audit trail; historical run records are deliberately **not** rewritten to point at the new IDs.

### Automatic purge in `runSet`

[...]

Issue #85, opened 2026-08-19
Epic: purging exclusions — remove data from backup history, Code42-style


bherila:
## Context

restic's `--exclude` is forward-only: it stops *new* snapshots from containing a path and does nothing to snapshots already written. When large build artifacts are swept into a repository before the exclude list covers them, deleting them locally does not remove them from the backup history — they remain in every snapshot taken while they existed.

Code42's exclusion list does both: adding a pattern removes matching files from the archive as well as from future backups. Restic Station should offer the same thing — a per-set list of patterns excluded going forward **and** rewritten out of existing snapshots, with the history cleanup happening on the next scheduled or manual run.

restic can do this (`restic rewrite --exclude <pat> --forget` creates replacement snapshots without the matching paths and drops the originals), but nothing in the codebase invokes `rewrite` today, and `prune` — the step that actually frees disk — exists only bundled into `forget --prune`.

## Semantics

Two lists on a `BackupSet`, deliberately distinct:

| field | effect on new snapshots | effect on existing snapshots |
|---|---|---|
| `excludes` (existing) | not backed up | **none** — unchanged behaviour |
| `purgeExcludes` (new) | not backed up | rewritten out, originals forgotten |

Existing configs are untouched: nothing anyone has already excluded is retroactively destroyed. Putting a pattern in `purgeExcludes` is the deliberate, warned act.

`purgeExcludes` is applied *in addition to* `excludes` when building `restic backup`, so the forward half is free. The retroactive half is a new engine phase that runs `rewrite` per destination.

**`rewrite` does not free space.** `rewrite --forget` unlinks the data; only `prune` reclaims it, and prune is repository-wide and expensive. Purge and prune stay separate operations — purge never auto-prunes.

## Design decisions

[...]

PR #97, opened 2026-08-20
Purging exclusions 2/5: read-only purge preview

### LINKED ISSUE #87: Purging exclusions 2/5: rewrite/prune argv, snapshot attribution, and a read-only purge preview
> **Before starting, read [#85 §Working agreement](https://github.com/bherila/restic-station/issues/85).** It carries the repo conventions, the `--json` envelope contract (new since this issue was written), what `local-ci.sh` does *not* cover, and the review process. Everything below assumes it.

> **Envelope delta for this issue.** `purge preview --json` must conform to `JSONRenderable`, emit through `CLIJSON.print` (so its payload lands under `data`), throw `CLIFailure` rather than calling `HelperExit.fail` on every reachable path including setup, and gain a row in `docs/cli-json.md`'s command matrix — the matrix is drift-tested. Snapshot IDs and paths belong in `data`; `details` is a fixed struct with no field that can hold a path, and that is deliberate.

> **Base:** `main` at `2f3c970` or later. #86 (1/5) is merged.

Part 2 of 5 of the purging-exclusions epic (#85). Depends on 1/5. Read #85 first — design decision 1 governs how snapshots are selected here.

Everything in this task is **read-only**. It adds the argv builders, the snapshot-attribution rule, and a preview command. Nothing it does can modify a repository.

### `Core/Sources/ResticStationCore/Restic/ResticCommand.swift`

Two new builders following the house pattern — operands only, fixed flag order, golden-tested against `docs/restic-cli.md`:

```
rewrite(repo:snapshotIDs:excludes:forget:dryRun:)
    -r R rewrite [--forget] [--dry-run] [--exclude p]... <id>...
prune(repo:dryRun:)
    -r R prune [--dry-run]
```

`rewrite` gets `precondition(!excludes.isEmpty)` and `precondition(!snapshotIDs.isEmpty)`. The second is load-bearing: it is what makes "rewrite every snapshot in the repository" unreachable by accident, mirroring `forget`'s empty-policy precondition.

`prune` as a standalone subcommand does not exist in the codebase today — all pruning is currently `forget --prune`.

### `Core/Sources/ResticStationCore/Engine/PurgePlan.swift` (new, pure, no I/O)

Attribution of `[Snapshot]` to a `BackupSet`, producing `PurgePlan { destinationId, matched: [Snapshot], unattributed: [Snapshot], patterns: [String] }`.

Attribution rule: `Snapshot.paths` is a subset of the union of the set's source paths across the shared config and every machine override, **and** `Snapshot.hostname` is one of the hostnames seen for the set. Reuses the existing `Snapshot` parser.

Keeping this pure is deliberate — it is the rule that decides what gets deleted, so it must be cheap to test exhaustively.

### Engine

`previewPurge(set:destination:) async -> PurgePlanResult` — set lock, reachability probe, `snapshots --json`, `PurgePlan`, then `rewrite --dry-run` (no `--forget`) over the matched IDs to learn which snapshots would actually change.

### Helper

`restic-station purge preview --set <uuid> [--dest <uuid>] [--json]` in a new `Helper/Sources/Commands/Purge.swift`. Human output names the affected snapshots, lists what it declined to attribute, and says plainly that space is not reclaimed until a prune.

## Fixture capture — do not guess

`rewrite` has **no `--json` mode**; its output is human text, like `copy`'s. Capture real restic output into `docs/fixtures/rewrite-dry-run.txt` and `docs/fixtures/rewrite-forget.txt` and copy them byte-identically into `Core/Tests/ResticStationCoreTests/Fixtures/`, per the existing convention. Confirm the exact per-snapshot wording against a real restic before writing any parser.

## Docs

`docs/restic-cli.md` gains a §rewrite and a §prune section with verified argv and exit codes, in the style of the existing §forget / §copy entries.

## Acceptance criteria

- [ ] `rewrite` argv matches `docs/restic-cli.md` §rewrite exactly, golden-tested.
- [ ] `ResticCommand.rewrite` cannot be constructed without at least one snapshot ID and one exclude pattern.
- [ ] A snapshot belonging to a different set sharing the same `repoURL` is never in `matched`, and appears in `unattributed` so it is visible rather than silently dropped.
- [ ] `purge preview` on a set with an empty `purgeExcludes` exits 0 and reports nothing to do.
- [ ] `purge preview` spawns no restic command that can modify the repository — asserted on the argv sequence, not just on behaviour.
- [ ] Fixtures are real captured restic output, not hand-written.
- [ ] `swift test` green for the root and Core packages on macOS and in the `swift:6.1` Linux container.

bherila:
Closes #87

## Summary

- add the fixed-order `rewrite` and standalone `prune` argv builders with non-empty ID/exclude preconditions
- attribute snapshots by source-path subset and hostname history across shared and machine override config
- add the read-only `purge preview` engine path and CLI, including the `--json` envelope contract
- capture and parse real restic 0.19.1 rewrite transcripts, with normative CLI docs and golden tests

The preview only invokes reachability, `snapshots --json`, and `rewrite --dry-run` with explicit matched snapshot IDs. It never invokes `rewrite --forget`, `prune`, or run-record mutation.

## Validation

- `./scripts/local-ci.sh` — passed all 8 checks
- `./scripts/headless-cli-test.sh` — passed
- `swift test --package-path Core --disable-sandbox` — 627 tests passed
- `swift test` — 89 tests passed
- real restic 0.19.1 scratch-repository fixture capture; docs/Core fixtures are byte-identical

Per #85, stop before #88: data deletion begins there.

### DIFF
--- a/Core/Sources/ResticStationCore/Engine/BackupEngine.swift
+++ b/Core/Sources/ResticStationCore/Engine/BackupEngine.swift
@@ -136,6 +136,11 @@ public final class BackupEngine: Sendable {
     private let reachability: Reachability
     private let now: @Sendable () -> Date
     private let uptime: @Sendable () -> TimeInterval
+    /// Raw shared-config source/hostname knowledge used by purge attribution.
+    /// Resolved configs deliberately strip machine overrides, so the helper
+    /// supplies these unions when it constructs the engine.
+    private let purgeSourcePaths: [UUID: Set<String>]
+    private let purgeHostnames: [UUID: Set<String>]
 
     public init(
         config: AppConfig,
@@ -146,7 +151,9 @@ public final class BackupEngine: Sendable {
         stateStore: StateStore,
         reachability: Reachability,
         now: @escaping @Sendable () -> Date = Date.init,
-        uptime: @escaping @Sendable () -> TimeInterval = { ProcessInfo.processInfo.systemUptime }
+        uptime: @escaping @Sendable () -> TimeInterval = { ProcessInfo.processInfo.systemUptime },
+        purgeSourcePaths: [UUID: Set<String>] = [:],
+        purgeHostnames: [UUID: Set<String>] = [:]
     ) {
         self.config = config
         self.paths = paths
@@ -157,6 +164,8 @@ public final class BackupEngine: Sendable {
         self.reachability = reachability
         self.now = now
         self.uptime = uptime
+        self.purgeSourcePaths = purgeSourcePaths
+        self.purgeHostnames = purgeHostnames
     }
 
     // MARK: - runSet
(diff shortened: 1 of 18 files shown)


bherila:
@codex review

PR #100, opened 2026-08-20
Fix standalone prune review follow-ups

bherila:
Follow-up to #99 for the final Codex review findings.

- keep busy `maintenance prune --dry-run` attempts out of run history
- bind reclaim confirmation to the exact previewed destination and require a new dry run after configuration changes
- normalize local paths before showing iCloud destructive-maintenance warnings

Validation:
- `swift test --package-path Core --filter BackupEngineTests`
- `xcodebuild -scheme 'Restic Station' test CODE_SIGNING_ALLOWED=NO -derivedDataPath \"$PWD/.build/review-dd\" -only-testing:'Restic StationTests/AppModelMachineOverrideTests'`\n- `./scripts/local-ci.sh`

PR #101, opened 2026-08-20
Restore reclaim confirmation on launch failure

bherila:
Follow-up to #100 review: restore a standalone-prune confirmation token only when `Process.run()` definitively fails before creating a child, and document the maintenance-prune JSON confirmation fields. Includes a regression test for the launch-failure path.


bherila:
@codex review

PR #102, opened 2026-08-20
Purging exclusions 5/5: remote SFTP prune

bherila:
Closes #90

Implements the final #85 child issue.

- adds opt-in SFTP remote maintenance with SSH argument quoting, `version --json` preflight, and no local fallback
- delivers repository passwords only through SSH stdin; argv, environment, and run logs exclude them
- adds destination-editor configuration plus a read-only remote-maintenance test
- preserves destructive preview tokens if SSH cannot launch

Validation:
- `./scripts/local-ci.sh` (real restic 0.19.1; disposable scratch repository only)
- `xcodebuild -project ResticStation.xcodeproj -scheme 'Restic Station' test CODE_SIGNING_ALLOWED=NO -derivedDataPath .build/issue-90-dd`

PR #99, opened 2026-08-20
Purging exclusions 4/5: standalone prune and Maintenance UI

bherila:
Fixes #89

- add retention-independent `maintenance prune` with dry-run and JSON envelopes
- keep standalone prune behind set locking, secret/reachability preflight, and mirror freshness
- add the shared Maintenance confirmation flow, iCloud materialization warning, and pending purge warning

Validated with `./scripts/local-ci.sh` (Linux cross-compile unavailable locally; CI covers Linux runtime/release checks).

PR #98, opened 2026-08-20
Purging exclusions 3/5: destructive purge apply

bherila:
Implements #88.

- Adds owner-only, expiring, single-use preview capabilities bound to machine/config/destinations/exact snapshot IDs.
- Applies `rewrite --forget` only after fresh attribution validation; records old-to-new snapshot mapping.
- Purges primary before mirror copies and skips a stale secondary when its purge cannot succeed.
- Adds `purge apply --set --preview-token [--json]`, JSON error classification, docs, UI run presentation, and regression coverage.

Validation:
- `./scripts/local-ci.sh` (all eight macOS/headless checks passed; local Linux cross-compile skipped because its SDK is not installed)
- real restic 0.19.1 rewrite --forget smoke test in a disposable `/tmp` repo

PR #103, opened 2026-08-21
Purge epic follow-ups: survive an early-exiting subprocess, close two fail-open edges

bherila:
Review follow-ups on epic #85 (#97, #98, #99, #100, #101, #102). Everything in
the epic is merged and CI on `main` is green; these are findings from reviewing
the merged result, not a re-litigation of it. Stacked so each can be judged
separately.

### 1. `DefaultProcessRunner` is killed by SIGPIPE when a child closes stdin first

The stdin channel added for the remote sftp prune wrote the repository password
with `FileHandle.write(_:)` and nothing masked SIGPIPE. A child that exits
before reading — `ssh` in BatchMode with a rejected host key does this in
milliseconds — closes the read end, and the write then takes down the whole
helper process, mid-destructive-operation, with no run record and no
diagnosis. `write(_:)` also raises an uncatchable ObjC exception rather than
throwing.

This is not theoretical: the new regression test kills the test binary with
**signal 13** against the unfixed runner.

The write also ran before the stdout/stderr readers started, so a child that
writes output before draining stdin would deadlock once the pipe buffer filled.
Today's only payload is a password, far under one buffer, but the reordering is
free and removes the class.

### 2. Two fail-*open* edges in the destructive-prune binding

- `remoteMaintenanceOperands()` split the repo URL on its first colon, which is
  only correct for the `sftp:user@host:/repository` shorthand. For restic's
  `sftp://user@host:port/repository` URL form that colon is the port, yielding
  sshTarget `//backup@example.com` and repoPath `2222/srv/repo`. Now only the
  shorthand is parsed; explicit operands still drive the URL form.
- `pruneConfirmationFingerprint` returned `""` on encode failure with a comment
  claiming that was "impossible-to-match". Both sides would compute `""` and
  compare **equal**, authorizing a prune against an unfingerprinted
  destination. Unreachable in practice, but the comment asserted a property the
  code did not have.

### 3. A crypto test gap, a misleading error message, and an undocumented rule

- `SHA256Digest` was covered only by the two shortest NIST vectors, both under
  one block — neither reaches the multi-block loop or the padding spill, though
  every fingerprint it computes is multi-block. Verified the implementation
  against `hashlib` first, then added 55/56/64/200-byte vectors as a guard.
- `PurgeApplyError.unavailable` covers secret storage, the token index, and a
  failed `snapshots` listing alike, but told the caller to check secret storage.
- `docs/scheduling.md` said "newly attributed snapshot ids" without ever
  defining attribution — the rule that decides what a purge may destroy. Now
  documented, including that subset membership can attribute one snapshot to
  two sets sharing a `repoURL`.

### Verification

`scripts/local-ci.sh` green (7/7 available; Linux cross-compile SDK not
installed locally). Both behavioural fixes red-checked against unfixed sources.
No `ci.yml` inline assertion touches the reworded message.

### DIFF
--- a/Core/Sources/ResticStationCore/Config/MachineConfig.swift
+++ b/Core/Sources/ResticStationCore/Config/MachineConfig.swift
@@ -1,5 +1,13 @@
 import Foundation
 
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#endif
+
 // MARK: - MachineConfig
 
 /// The host-local half of the configuration (`machine.json`).
@@ -112,6 +120,37 @@ public enum MachineIdentity {
     /// nothing — a fresh lowercased UUID (which is itself a valid slug).
     public static func generate(hostName: String = ProcessInfo.processInfo.hostName) -> String {
         slugify(hostName) ?? UUID().uuidString.lowercased()
+    }
+
+    /// The kernel hostname, i.e. `gethostname(2)` — which is what Go's
+    /// `os.Hostname` returns and therefore what **restic stamps into every
+    /// snapshot**.
+    ///
+    /// This is not the same string as ``generate(hostName:)`` uses.
+    /// `ProcessInfo.hostName` is the reverse-DNS name and on macOS usually
+    /// carries a `.local` suffix the kernel name does not: on a stock Mac,
+    /// `ProcessInfo.hostName` is `bens-laptop.local` (→ machineId
+    /// `bens-laptop-local`) while restic records `Bens-Laptop` (→
+    /// `bens-laptop`). Purge attribution compares against snapshot
+    /// hostnames, so it must know both or it matches nothing this machine
+    /// ever wrote.
+    public static func kernelHostName() -> String? {
+        var buffer = [CChar](repeating: 0, count: 256)
+        guard gethostname(&buffer, buffer.count) == 0 else { return nil }
+        let name = String(cString: buffer)
+        return name.isEmpty ? nil : name
+    }
+
+    /// Every slug that may legitimately identify **this** host in a restic
+    /// snapshot: the configured `machineId` and the kernel hostname restic
+    /// actually records. Callers attributing snapshots must use this rather
+    /// than `machineId` alone.
+    public static func localHostnameSlugs(machineId: String) -> Set<String> {
+        var slugs: Set<String> = [machineId]
+        if let kernel = kernelHostName(), let slug = slugify(kernel) {
+            slugs.insert(slug)
+        }
+        return slugs
     }
 }
 
--- a/Core/Sources/ResticStationCore/Config/Models.swift
+++ b/Core/Sources/ResticStationCore/Config/Models.swift
@@ -359,7 +359,14 @@ public extension Destination {
         guard kind == .sftp else { return nil }
         let configured = remoteMaintenance ?? RemoteMaintenance()
         let raw = String(repoURL.dropFirst("sftp:".count))
-        let split = raw.firstIndex(of: ":")
+        // Only the `sftp:user@host:/repository` shorthand can be split on its
+        // first colon. In restic's `sftp://user@host:port/repository` URL form
+        // that colon introduces the port, so the same split would aim ssh at
+        // "//user@host" and hand restic "port/repository". Decline to infer
+        // rather than build a nonsense remote command; explicit `sshTarget`
+        // and `remoteRepoPath` still drive that form.
+        let shorthand = raw.hasPrefix("//") ? nil : raw
+        let split = shorthand.flatMap { $0.firstIndex(of: ":") }
         let parsedTarget = split.map { String(raw[..<$0]) }
         let parsedPath = split.map { String(raw[raw.index(after: $0)...]) }
         guard let sshTarget = configured.sshTarget ?? parsedTarget, !sshTarget.isEmpty,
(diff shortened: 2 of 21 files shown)


bherila:
@codex review

bherila:
@codex review

bherila:
Both findings addressed in `8ee48a5`.

**Child signal dispositions** — real concern, and the fix now pins it rather than assuming it. Your stated repro does not hold here: `/bin/bash -c 'kill -PIPE $$; echo survived'` launched through Foundation's `Process` still dies with signal 13, because `Process` resets child dispositions. That is undocumented, so I added `childrenKeepTheDefaultSIGPIPEDisposition`, which spawns *after* the stdin path has run and asserts the child still takes signal 13 — and it runs on the Linux CI container, i.e. the platform I cannot check locally is the one proving the invariant.

Your suggested alternative — mask SIGPIPE only around the parent's write — I tried first, and it does not work. With SIGPIPE blocked via `pthread_sigmask` on the writing thread and the raw `write` correctly returning `EPIPE`, the process still died: Foundation raises the signal on an internal queue thread, which a per-thread mask cannot cover. That is now recorded in the comment so the next reader doesn't repeat the experiment.

**Attribution docs** — straightforwardly wrong as written, thank you. `HelperContext` always inserts its own `machineId` (`HelperContext.swift:179`), so the local machine's snapshots *are* attributable with no `machines` entry; what the absence of an entry protects is another host's snapshots. Reworded. This section is normative about what a purge may destroy, so an over-broad safety claim was the worst kind of error to leave in it.

bherila:
Linux CI proved you right, and proved my previous reply wrong on both counts. Correcting the record:

**1. The leak is real — on Linux.** Against the process-wide `signal(SIGPIPE, SIG_IGN)`, the test I added to "pin" the invariant *fails* in the `swift:6.1` container: the child returns exit 0 and prints `survived` instead of dying with signal 13. swift-corelibs-foundation does not reset child signal dispositions; macOS's Foundation does, which is why my repro of your repro came back clean. My "your reproduction does not hold here" was true only of macOS, and macOS was the platform hiding the bug.

**2. My reason for rejecting your suggested fix was also wrong.** I claimed a scoped window could not work because Foundation raised the signal on an internal queue thread. A `sigaction` handler shows SIGPIPE is raised **synchronously on the writing thread** — my earlier `pthread_sigmask` attempt failed on a drain bug of mine, and I generalized from a failed experiment instead of measuring it.

`9472923` does what you suggested: the ignore window covers exactly the write, and a `spawnLock` serializes it against `process.run()`, so the disposition is `SIG_DFL` at every spawn on both platforms and no child can be created while it is ignored. `linux` is now green on that sha.

The test stays, since it is the only thing that distinguishes the two implementations — `local-ci.sh` was green on this branch through both.

bherila:
## Fable review complete — four reviewers across engine, capability/security, helper CLI, and the app layer

The most important finding is that **the purge feature did nothing at all on this machine**, silently, while reporting success.

Attribution compares a snapshot's hostname against `machineId`, which is slugified from `ProcessInfo.hostName` (reverse-DNS). restic stamps the kernel name from `gethostname(2)`. On a stock Mac those differ — `bens-laptop.local` → `bens-laptop-local` vs restic's `Bens-Laptop` → `bens-laptop` — so **zero** snapshots were attributable. Verified end to end with the real types on this host: `matched=0, unattributed=1` for a snapshot the machine had just written.

The failure was invisible: `runAutomaticPurge` got an empty plan, took the empty-match branch, advanced `appliedPurgeExcludes`, and reported success. Permanently — the rewrite never ran again on any later run. Every historical snapshot kept exactly the content the user asked to have purged. For a feature whose purpose is removing data from history, that is the worst available outcome, and no existing test caught it because the engine tests inject `purgeHostnames: ["example-mac.local"]` to match their fixture.

Fixed in `ace9a51`, with the empty-match branch no longer advancing the watermark when snapshots were *declined* (the amplifier that made it permanent), and a regression test asserting against live system APIs rather than a fixture.

Also fixed here:
- **P2, reproduced** — `purge apply --preview-token` rejected any token starting with `-` (~1 in 64) in the exact form `purge preview` prints, burning a reviewed 15-minute token on an unavoidable usage error.
- **P2** — a partitioned remote prune hung the helper forever holding the set lock, silently stopping every scheduled backup for that set. `ConnectTimeout` bounds only the handshake; added ServerAlive keepalive. Not a wall-clock timeout — a legitimate prune runs for hours, and only a keepalive distinguishes slow from gone.
- An expired reclaim binding reporting `operation_not_allowed` / "configuration changed", which was false and disagreed with `purge apply`.
- Token-store contention reported as non-retryable `internal_error`.
- `docs/cli-json.md` drift ("Eleven commands" vs 14; "two codes have no producer" when both now do), and `jsonCapable` omitting both destructive purge commands.

Filed rather than fixed here, to keep this PR scoped: #104 (retention apply has no fail-closed binding — pre-existing, and the asymmetry with reclaim is the interesting part), #105 (Maintenance UI stale-state), #106 (SSH TOFU + capability values in argv + token-store hardening), #107 (schedule-state lost update), #108 (helper's own stdout/stderr SIGPIPE).

What the reviewers verified clean is worth recording too: no path can reach `rewrite` without explicit snapshot IDs; apply scope exactly equals the reviewed plan; SHA-256 matches CryptoKit across padding boundaries; `ShellQuote` round-trips injection attempts byte-exactly through sh/bash/zsh; the password never reaches argv, env, run records, or envelopes; and the new `spawnLock` has no re-entrancy or unlocked-spawn holes.

PR #125, opened 2026-08-24
Require jq for integration JSON assertions

### LINKED ISSUE #123: integration-test.sh's no-jq python fallback emits Python literals, not JSON
`scripts/integration-test.sh` supports a `python3` fallback for JSON queries when `jq` is absent (`preflight_jq`, `idx_select`). That path is broken and has been for some time — it is simply never exercised, because CI installs `jq` on every job that runs this script.

## Reproduction

Force the fallback (both `jq` and `brew` must be unavailable, or patch `preflight_jq` to `HAVE_JQ=0`), then run the script:

```
ASSERTION FAILED at step: run3 (unplug secondary)
Reason: expected secondary reachable:false, got 'False'
```

## Cause

The fallback prints Python values with `str()` rather than serializing to JSON, so booleans come out as Python's `True`/`False` instead of JSON's `true`/`false`. Any assertion comparing against a JSON literal fails.

The failure is at the first boolean comparison; there may be more of the same class further in, plus the `None` vs `null` equivalent. Worth auditing the whole fallback rather than patching the one field.

## Why it matters

Low urgency — CI is unaffected, since `.github/workflows/ci.yml` installs `jq` for the `macos` and `linux-integration` jobs. But the fallback exists so the script runs on a host without `jq`, and it does not, so it is worse than not having it: a contributor on such a host gets a confusing failure in an unrelated step rather than a clear "install jq".

Either fix the fallback and add a CI job that exercises it, or delete it and make `jq` a hard prerequisite with a clear preflight error. A fallback that is never run will rot again.

## Note

Found while verifying a whitespace-tolerant `trigger` match in #122 — that fix is correct (both `jq -c`'s `"trigger":"scheduled"` and `json.dumps`'s `"trigger": "scheduled"` match), but it could not be confirmed end-to-end on the fallback because the script fails earlier for this unrelated reason.

bherila:
## What changed

Resolves #123 by removing the unexercised Python JSON-query fallback from `scripts/integration-test.sh`. The script now requires `jq` and fails before creating test state with an actionable prerequisite error. CI already installs `jq`; no dormant fallback remains. Updated the normative T19 task contract in the same change.

## Verification

- `bash -n scripts/integration-test.sh`
- Isolated PATH with a fake `restic` and no `jq`: exits 1 with the exact prerequisite error and no fallback claim.
- `scripts/integration-test.sh` exits 0 locally by its documented no-restic local skip.
- `git diff --check`

No test assertions were added, so a mutation check is not applicable.

## Review anchor

- Base SHA: `77a20f8b9870ad75288875156ddb18805c015d4b`
- Head SHA: `79345a648d7769c37056171d2762e4573bb9f208`
- Reviewed at this pair: pending

## Checklist

- [x] Documentation reconciled with the behavior change
- [x] No application or destructive-path behavior changed
- [x] Clear no-`jq` failure was exercised
- [x] `git diff --check` passes

### DIFF
--- a/docs/tasks/T19-integration-test.md
+++ b/docs/tasks/T19-integration-test.md
@@ -11,7 +11,7 @@
   shebang's argument string on whitespace. Linux passes the remainder as a *single* argument, so bash
   consumes the script path as `-o`'s option name and exits with "invalid option name" before running
   anything. T29 hit this the first time the script ran on Linux — do not reintroduce it.) Structure:
-  - Preflight: restic on PATH (else skip with exit 0 + notice when `CI` unset; hard fail in CI), `jq` available (brew or bundled fallback via `python3 -c`).
+  - Preflight: restic on PATH (else skip with exit 0 + notice when `CI` unset; hard fail in CI), and `jq` required with a clear failure before the script creates test state. The script has no unexercised JSON-parser fallback.
   - Workspace: `mktemp -d`; `RESTIC_STATION_DATA_DIR` exported; `trap` cleans workspace AND keychain items AND kills stray helpers.
   - Keychain seeding: two fixed test UUIDs; `security add-generic-password -s restic-station -a <uuid> -w test-password -T /usr/bin/security` (delete-first for idempotency). On CI, create/unlock a temporary keychain if `security show-keychain-info` fails (document the runner behavior encountered).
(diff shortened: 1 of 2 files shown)


bherila:
@codex review

Review anchor:
- Base SHA: `77a20f8b9870ad75288875156ddb18805c015d4b`
- Head SHA: `79345a648d7769c37056171d2762e4573bb9f208`

PR #135, opened 2026-08-26
Add a fault-injection crash matrix for run-record durability, fix a tail-repair bug it found

bherila:
## What changed

Two commits. **Commit 1** adds a fault-injection harness over the `RunStoreFileOperations` seam #129 introduced: `FaultInjectingFileOperations` can crash at any seam-call boundary (subsequent operations no-op — the in-process analog of process death), fail or EINTR any fsync by ordinal, tear a write at a byte offset, and record an ordering trace. `RunStoreCrashDurabilityTests` runs parameterized crash matrices over every step boundary of `begin`, `markDestructiveLaunchAuthorized`, `finish` (destructive and non-destructive), and `recoverInterrupted` itself, remounting a fresh store each time and asserting the recovered view is consistent; plus torn appends at offsets including mid-UTF-8-scalar, EINTR at every fsync site, and direct trace assertions of the documented write→fsync→rename→dir-fsync and marker-before-append orderings. Each matrix's largest crash point asserts flow completion, so a flow that grows steps fails coverage loudly. Documented in `docs/testing.md` with the model's limits (page-cache loss and mkdir/remove boundaries not representable). Production behavior untouched by this commit.

**Commit 2** fixes a real bug the matrix found: `readIndexEntriesRepairingTail` reads only with `pread`, so its descriptor's offset was still 0 when it wrote the terminating newline for a complete newline-less final record — the newline overwrote the first byte of `index.jsonl`, **destroying the oldest run record** while leaving the tail unterminated. Fix: `O_APPEND` on the repair descriptor. (`appendIndexEntry`'s descriptor already carried `O_APPEND` and was safe.)

Motivation: #129 drew 16 review findings because fsync-ordering/torn-write bugs were visible only to reviewers, never to tests (the ancestor-fsync gap was re-flagged three times). This harness makes that class locally testable — and found a live data-loss bug on its first run.

Per the in-flight-#130 conflict constraint, `ConfigStore` shares none of these seams and is untouched; seaming it is a follow-up after #130 lands.

## Review anchor

An independent review applies to an exact `(base SHA, head SHA)` pair. Changing
the base can change the diff even when the head commit is identical, so these
must be filled in and **re-stated after any rebase, force-push, or retarget** —
a review against a superseded pair does not carry forward.

- Base SHA: `213da09f4229f7bdbf2a5df462ea9e45079999e8`
- Head SHA: `e490c911afa9aa9cd68fbaf11a934203f4072933`
- Reviewed at this pair: Codex clean pass (thumbs-up), zero findings

## Checklist

- [x] Spec docs in `docs/` updated in this PR if behavior changed (they are normative) — `docs/testing.md` documents the harness; the tail-repair fix restores the behavior `data-model.md` already specifies (repair terminates the tail; no record loss)
- [x] `project.yml` edited rather than `ResticStation.xcodeproj` (it is generated) — neither touched
- [x] `Core/` and `Helper/` still build and test on Linux — no Darwin-only API unguarded (Musl import guards in tests); local `swift build`/`swift test` root 115/21 + Core 816/91 green
- [x] Invariants preserved — no acknowledged run record may be lost: previously violated by the repair path, now enforced and pinned by test
- [x] Destructive-path change? The run index feeds the destructive audit. Failure direction if the fix is wrong: recovery refuses/fails loudly under the crash matrix (fails closed); the *old* code silently destroyed the oldest record (failed open) — this PR moves the failure to the recoverable side.
- [x] Behavior change? Only the bug fix; blast radius swept — no doc, envelope, or copy describes the corrupted-head behavior
- [x] New/changed assertions red-checked — `recoveryTerminatesNewlinelessTailAtTheEnd` fails against the pre-fix flags (2 issues: head overwritten, tail unterminated); matrix coverage checks fail when a flow grows steps
- [x] New or extended safety invariant? Threat model stated: crash at every seam-call boundary of every publication flow, torn writes at arbitrary offsets (incl. mid-scalar), EINTR at every fsync, recovery-during-recovery idempotence, dead-pid `.running` records

https://claude.ai/code/session_01KtczdRBghL2e9CxbXbrNMC

### DIFF
--- a/Core/Sources/ResticStationCore/Runs/RunStore.swift
+++ b/Core/Sources/ResticStationCore/Runs/RunStore.swift
@@ -949,10 +949,15 @@ public struct RunStore: Sendable {
 
         let runsDirectoryFD = try openDirectory(paths.runsDir)
         defer { closeFileDescriptor(runsDirectoryFD) }
+        // `O_APPEND` is load-bearing: the tail repair reads only with
+        // `pread`, which leaves the file offset at 0, so the terminating
+        // newline written for a complete newline-less final record would
+        // otherwise land on the first byte of the index — destroying the
+        // oldest record instead of finishing the newest one.
         let fd = fileOperations.openAt(
             runsDirectoryFD,
             paths.runsIndexFile.lastPathComponent,
-            O_RDWR | O_CLOEXEC | O_NOFOLLOW,
+            O_RDWR | O_APPEND | O_CLOEXEC | O_NOFOLLOW,
             0
         )
         guard fd >= 0 else {
--- a/Core/Tests/ResticStationCoreTests/Runs/RunStoreCrashDurabilityTests.swift
+++ b/Core/Tests/ResticStationCoreTests/Runs/RunStoreCrashDurabilityTests.swift
@@ -410,10 +410,12 @@ import Musl
         }
 
         // `lineLength - 1` keeps the whole JSON record and tears only the
-        // newline — the complete-newline-less tail covered by
-        // `recoveryTerminatesNewlinelessTailAtTheEnd`; excluded here so the
-        // torn-fragment matrix stays focused on strict prefixes.
-        for offset in stride(from: 1, to: lineLength, by: 3) where offset != lineLength - 1 {
+        // newline — the complete-newline-less repair path also pinned by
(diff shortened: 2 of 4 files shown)


bherila:
@codex review

Issue #116, opened 2026-08-21
Intermittent posix_spawn EFAULT in the Core test suite under parallel execution


bherila:
Measured, pre-existing, and not yet root-caused. Filed so it is tracked rather than rediscovered.

## Symptom

`Process.run()` intermittently fails with `NSPOSIXErrorDomain Code=14 "Bad address"`, surfacing as `ProcessRunnerError.launchFailed`. It has been observed in `DefaultProcessRunnerTests` but the failing test varies — whichever spawn happens to lose.

## Measurements

| Configuration | EFAULT rate |
|---|---|
| `main` (84a8a2e), `swift test --package-path Core` | **1 / 25** |
| `agent/review-followups-104-108`, same command | **1 / 25** |
| Branch, `--no-parallel` | 0 / 6 |

The two branches are indistinguishable, so this is **not** introduced by #109. Smaller samples are badly misleading here: separate 10–14 run batches produced 0/10, 1/12, 1/14, 3/10 and 3/14 for configurations that later measured identically. Anyone investigating should use ≥25 runs per configuration before concluding anything.

`--no-parallel` showing 0/6 is suggestive of a concurrency interaction but is far too small a sample to rely on.

## Why it matters

Roughly 4% per suite run is enough to make CI flaky, and — more importantly — the same code path spawns restic concurrently across destinations in production. A spurious `launchFailed` there is reported as a failed operation, so a backup or destructive maintenance step could fail for no real reason.

## Suggested investigation

- Establish the rate properly (≥25 runs per configuration) before and after any change.
- Determine whether the failure is in `posix_spawn` itself or in Foundation's marshalling around it; capture the actual `errno` at the syscall rather than the mapped `NSError`.
- Check whether it correlates with total concurrent spawn count, with thread-pool pressure, or with signal-disposition changes concurrent with a spawn.
- Consider serializing spawns if the cause turns out to be Foundation-side, but measure rather than assume — serializing was tried during #109 investigation and did not clearly help.

## Related
Not to be confused with #114 (pipe-driven timeout and process-group termination), which is a separate defect in the same file.

bherila:
Measurement campaign run 2026-08-23 (macOS 26.6.1, M-series arm64, 10 cores, Swift 6.3.3, repo @ `e813207c`). Raw logs and a machine-readable table were produced locally; the numbers below are the durable record.

| Configuration | Runs | EFAULT failures |
|---|---|---|
[...]

PR #146, opened 2026-08-27
Reject "." and ".." components in configured source paths

### LINKED ISSUE #143: Reject '.' and '..' components in configured source paths at validate time
From the pre-merge review of #139.

`PurgePlan.normalizedForComparison`
(`Core/Sources/ResticStationCore/Engine/PurgePlan.swift`) collapses repeated `/`
and strips a trailing `/`, but **deliberately leaves `.` and `..` alone** — the
type has no filesystem dependency, and resolving `..` lexically is wrong in the
presence of symlinks.

restic, however, lexically cleans the paths it records in snapshots. So a
configured source containing a `.` or `..` component produces snapshot paths that
never compare equal to the configured source, and every snapshot from that set
becomes **unattributable**.

## Behavior today

Purge fails closed and loudly — the plan matches nothing while snapshots were
declined, which #139 made an explicit apply failure rather than a silent success.
So this is safe, just badly timed: the operator learns about it at purge time,
from a message about attribution, rather than at config time.

## Ask

Add an `AppConfig.validate()` rule rejecting `.` and `..` components in source
paths, with the existing negative-test treatment required by
`docs/data-model.md` §Invariants. Surfacing it at `config validate` turns a
confusing purge-time refusal into an obvious configuration error.

Worth confirming first whether any real config in the fleet uses such a path,
since this would newly refuse it.

bherila:
Follow-up #143 from the #139 pre-merge review.

## Why

restic records the lexically cleaned path in a snapshot. `PurgePlan`
deliberately does **not** resolve `.`/`..` — it has no filesystem access, and
resolving `..` lexically is wrong across a symlink. So the configured path and
the recorded path never compare equal, and every snapshot from such a source is
unattributable: the plan matches nothing while snapshots were declined, which
#139 made an explicit apply failure.

The behavior is already **safe** — it fails closed, loudly. It is just
diagnosed at purge time, from a message about attribution, for what is plainly
a configuration error. This moves the diagnosis to `config validate`.

## Scope

- Applied to **override** sources too. The override path re-validates what it
  replaces, so leaving it out would make a machine override a way around the
  rule.
- **Empty components stay legal.** `PurgePlan` normalizes repeated and trailing
  separators on both sides, so `/srv//data` and `/srv/data/` already attribute
  correctly; refusing them would reject working configurations.
- **Dotfiles are not `.` components.** `/home/me/.config`, `/srv/..data`,
  `/srv/data.` are all accepted — asserted, because that is the obvious way to
  get this rule wrong.

## Blast radius

`ConfigError`'s classification switches are exhaustive on purpose, so three
places failed to compile until each was given the new cases — the mechanism
from #134 doing its job:

| Site | Given |
|---|---|
| `CLIError.classify` | collapsed to `config_invalid`, like every sibling |
| `FailureClassificationTableTests` | table rows |
| `SetsCopy.fieldMessage` | inline messages in that screen's voice |

Note the GUI's source picker returns resolved paths, so this rule only bites a
hand-edited `config.json`.

## Verification

Red-checked — removing either check fails the new tests with *"an error was
expected but none was thrown"*, across all four path shapes and the override.

`./scripts/local-ci.sh` green, including the Linux musl cross-compile. **One
run reported a `Core package tests` failure** that did not reproduce in two
subsequent full runs or a standalone 911-test run; that is consistent with the
known intermittent in #116, not with this change. Flagging it rather than
quietly re-running until green.

## Review anchor

- Base: `f61abdc04fee43a4047ad079b8e33ec39db11d9c`
- Head: `fc32dff0ca5a518811cb02869743c59bade4845a`

Closes #143

https://claude.ai/code/session_01EstNhm3JVkixiiJEQtxLtB

### DIFF
--- a/App/Sources/Views/Sets/SetsCopy.swift
+++ b/App/Sources/Views/Sets/SetsCopy.swift
@@ -207,6 +207,13 @@ enum SetsCopy {
             return (.sources, "Add at least one folder or file to back up.")
         case .relativeSourcePath(_, let path):
             return (.sources, "Sources must be absolute paths — “\(path)” is not.")
+        case .nonCanonicalSourcePath(_, let path):
+            return (
+                .sources,
+                "“\(path)” contains “.” or “..”. Use the folder's real path instead — restic "
+                    + "stores the resolved path, so backups from this source could not be matched "
+                    + "back to it."
+            )
         case .emptyPurgeExcludePattern:
             return (
                 .purgeExcludes,
@@ -235,6 +242,13 @@ enum SetsCopy {
             return (
                 .sources,
                 "Sources must be absolute paths — “\(path)”, set for machine “\(machineId)” in config.json, is not."
+            )
+        case .nonCanonicalOverrideSourcePath(_, let machineId, let path):
+            return (
+                .sources,
+                "“\(path)”, set for machine “\(machineId)” in config.json, contains “.” or “..”. "
+                    + "Use the folder's real path instead — restic stores the resolved path, so "
+                    + "backups from this source could not be matched back to it."
             )
         case .notExactlyOnePrimaryDestinationForMachine(_, let machineId, let count):
             if count == 0 {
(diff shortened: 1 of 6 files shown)


bherila:
@codex review

